PR #1, opened 2018-08-15
Add accuracy tests to the parser

SamDepardieu:
This PR aims to bring some tests to check the accuracy of the parser.
From the initial PR:
- There is a python file, separate from main, that runs all tests and produces numbers for each step.
- The code uses files with the same format as in AWS and adds only the manually labeled files that contain the ground truth.
- The tests are independent so each test tests one step of the process assuming everything worked perfectly before.
- The tests output a percentage for each step, e.g. 80% accuracy with some definition of accuracy that combines the labels that exist in the files for example accuracy=percentage of good and perfect over all


SamDepardieu:
[Add Liz's work to the new parser repository](https://trello.com/c/6zzRQdEs/394-add-lizs-work-to-the-new-parser-repository)

lizgzil:
I think it is literally just deleting the `test_structure` function

SamDepardieu:
@lizgzil I updated this, can you review again? 😄

PR #2, opened 2018-08-16
Add a Contributing file

SamDepardieu:
Add a contribution file to guide future contributors through our values, our workflow and our rules.
At the moment, it is a copy from this file -> https://www.contributor-covenant.org/version/1/4/code-of-conduct
Pretty much everything is in it, but we should make sure everything we want lives in this file.
This file should be discussed to ensure that we all agree with what is in it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,16 +1,15 @@
 # Contributing
 
-:tada: First of all, welcome to this repository, and thanks for your interest, wether you want to contribute
+:tada: First of all, welcome to this repository, and thanks for your interest, whether you want to contribute
 or simply are interested in opening an issue.
 
-When contributing to this repository, please first discuss the change you wish to make via issue,
-email, or any other method with the owners of this repository before making a change.
+When contributing to this repository, please first discuss the change you wish to make by creating an issue before making a change.
 
 Please note we have a code of conduct, please follow it in all your interactions with the project.
 
 ## Pull Request Process
 
-1. Before pushing your code, please make sure to have it tested and covered by some unit tests.
+1. Before pushing your code, please make sure to have it covered by some unit tests. Also make sure that the code passes both your tests and the other tests.
 
 2. Use our pull request template to submit any changes to the code, even minor.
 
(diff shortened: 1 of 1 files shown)


SamDepardieu:
[Create a contribution template for the parser](https://trello.com/c/vhhe1Cah/367-create-a-contribution-template-for-the-parser)

CONTRIBUTING.md
```diff
@@ -0,0 +1,97 @@
...
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+### Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at [INSERT EMAIL ADDRESS]. All
```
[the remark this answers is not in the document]
SamDepardieu: I don't know actually 🤔 
I'm removing this bit, we'll come with something later.

CONTRIBUTING.md
```diff
@@ -0,0 +1,96 @@
+# Contributing
+
+:tada: First of all, welcome to this repository, and thanks for your interest, wether you want to contribute
```
nsorros: whether instead of wether

CONTRIBUTING.md
```diff
@@ -0,0 +1,96 @@
...
+or simply are interested in opening an issue.
+
+When contributing to this repository, please first discuss the change you wish to make via issue,
```
nsorros: via an issue instead of via issue

CONTRIBUTING.md
```diff
@@ -0,0 +1,96 @@
...
+email, or any other method with the owners of this repository before making a change.
```
nsorros: or any other method? i assume we need to be specific about the method of communicating

CONTRIBUTING.md
```diff
@@ -0,0 +1,96 @@
...
+
+Please note we have a code of conduct, please follow it in all your interactions with the project.
+
+## Pull Request Process
+
+1. Before pushing your code, please make sure to have it tested and covered by some unit tests.
```
nsorros: tests need to pass as well

CONTRIBUTING.md
SamDepardieu: Sounds right. I think issue conversations are the best way for us. Any thoughts about that?

SamDepardieu:
@lizgzil @jdkram @nsorros This is ready for another round! 👌 ☕️

PR #10, opened 2018-10-19
Add a gitignore file

SamDepardieu:
# Description

Add a gitignore file, because we need a gitignore file

## Type of change

- [x] :sparkles: New feature


.gitignore
```diff
@@ -0,0 +1,10 @@
+*.pyc
+*__pycache__
```
hblanks: py27 and py3k!

PR #9, opened 2018-10-01
[WiP] Easy mode api

SamDepardieu:
# Description

This pull request aims to create a way for the user to easily use this code base, either by creating a service from it (using Docker) or by calling it as a command line tool.

## Type of change

Please delete options that are not relevant.

- [ ] :bug: Bug fix (Add `Fix #(issue)` to your PR)
- [x] :sparkles: New feature
- [ ] :fire: Breaking change
- [ ] :memo: Documentation update

# How Has This Been Tested?

This is still in progress 🚧 

# Checklist:

- [x] My code follows the style guidelines of this project (pep8 AND pyflakes)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] If needed, I changed related parts of the documentation
- [ ] I included tests in my PR
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] If my PR aims to fix an issue, I referenced it using `#(issue)`

PR #14, opened 2018-10-26
Utils structuring

SamDepardieu:
Second part of #9 

Mostly about moving files/ transforming them into independant modules


main.py
```diff
@@ -1,163 +1,65 @@
-import pandas as pd
-# import numpy as np
 import time
-import utils
-from models import Document, Organisation, Reference, DatabaseEngine
-from io import StringIO
+import os
+from utils import (FileManager, FuzzyMatcher, process_reference_section,
+                   Predicter)
```
hblanks: Some may prefer one import per line as it can make additions/deletions easier to diff. Don't change, just FYI.

main.py
```diff
@@ -166,21 +68,43 @@ def save_to_database(documents, references, session):
...
-        str(len(reference_components)),
-        " is ", str(total)
+        "Time taken to predict and match for %s is %s",
+        str(len(splited_references)),
+        str(total)
     )
+
+
+if __name__ == '__main__':
+
```
hblanks: Even for internal tools, sometimes it's handy to have things here like:

```
from argparse import ArgumentParser
parser = ArgumentParser(description=__doc__.strip())
parser.parse_args()
```

So that `python path/to/script.py -h` prints help. Again, though, no need to change.

models.py
```diff
@@ -49,8 +53,71 @@ class Reference(Base):
...
-    def get_session(url):
+    def _get_session(self, url):
         engine = create_engine(url)
         Base.metadata.bind = engine
         DBSession = sessionmaker(bind=engine)
         return DBSession()
+
+    def __init__(self):
+        self.meta = Base.metadata
+        self.session = self._get_session(settings.RDS_URL)
```
hblanks: FYI that at some point, having one session per instance with no explicit teardown may not be the best idea; a more viable pattern sometimes is to keep the session decoupled from business logic like `save_to_database()`, usually by having the session be the first argument to the business logic function. I.e.:

```
def save_to_database(session, documents, references):
    ...
```

(so, the session dependency is injected in, rather than sourced from a common object)

As before, we shouldn't change this as part of this cleanup, just an FYI.

main.py
hblanks: Also, not to change, but modules named `utils` (or `misc`) may be ripe for cleaner names eventually. Not today, though.

PR #23, opened 2018-12-10
Fix the reference parser performances issues

SamDepardieu:
# Description

[This PR looks big, but it's mostly due to switching back the Predictor class to a module with functions]

The reference parser had big ussues dealing with big files (e.g. the who scraping results, about 1.5Go). This PR aims to fix these issues by:

- Using multiprocessing to predict the references components and references
- Using a tempfile to read the scraping result file from S3
- Parsing the json file prior to its treatment in memory to remove big and useless chunks from it (e.g. the PDF full text)
- Indexing the database (manually done - see issue on datalabs infra for more details)
- Using an UPSERT statement instead of a SELECT and an INSERT (twice less requests to the database)

## Type of change

- [x] ⚡️ Performance update

# How Has This Been Tested?

This has been tested by running pultiple tests both locally and in the cloud in order to benchmark performances before and after these changes.

# Checklist:

- [x] I included tests in my PR - Irrelevant
- [x] If my PR aims to fix an issue, I referenced it using `#(issue)` - Irrelevant

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,4 +14,3 @@ six==1.11.0
 sklearn==0.0
 SQLAlchemy==1.2.12
 urllib3==1.23
-jsonlines==1.2.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,9 +8,10 @@ psycopg2-binary==2.7.5
 python-dateutil==2.7.3
 pytz==2018.5
 s3transfer==0.1.13
-scikit-learn==0.20.0
+scikit-learn==0.19.2
 scipy==1.1.0
 six==1.11.0
 sklearn==0.0
 SQLAlchemy==1.2.12
 urllib3==1.23
+jsonlines==1.2.0
--- a/utils/file_manager.py
+++ b/utils/file_manager.py
@@ -1,8 +1,9 @@
 import pandas as pd
 import os
 import pickle
+import tempfile
+import jsonlines
 from .s3 import S3
-from io import StringIO
 from settings import settings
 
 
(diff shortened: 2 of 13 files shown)


utils/file_manager.py
```diff
@@ -20,21 +21,52 @@ def __init__(self, mode='LOCAL'):
...
+                    "sections",
+                    "uri"
+                ]
+                json_file = []
+                for obj in jsonlines.Reader(tf):
+                    new_row = {}
+                    for column in columns_to_keep:
+                        new_row[column] = obj[column]
+                    json_file.append(new_row)
+            return pd.DataFrame(json_file)
```
hblanks: I think this could be done without jsonlines and with with 1/2 the memory:

```python

SCRAPING_COLUMNS = ('title', 'hash', 'sections', 'uri')

def to_row(self, line, lineno):
    try:
        return {
            k: v for k, v in json.loads(line).items()
            if k in SCRAPING_COLUMNS
        }
    except Exception as e:
        self.logger.error("Error on line %d: %s", lineno, e)
        raise



def get_scraping_results(...):
    ...
        with tempfile.TemporaryFile() as tf:
            ...
            tf.seek(0)
            rows = (self.to_row(line) for line in tf)
            return pd.DataFrame(rows)

```

It may be we wise to also ad docstrings for `_get_from_local()` vs `_get_from_s3()`, etc. It's not clear when `get_scraping_results()` is called and when these other methods are called.

utils/s3.py
```diff
@@ -15,11 +17,11 @@ def _get_last_modified_file_key(self, prefix):
                 Prefix=prefix
             ).get('Contents', [])
         except ClientError:
-            print('Could not connect to s3 bucket.')
+            self.logger.info('Could not connect to s3 bucket.')
             return ''
 
         if not objs:
-            print('Could not get last result file.')
+            self.logger.info('Could not get last result file.')
```
hblanks: Thank you for all these logging statements!

utils/file_manager.py
```diff
@@ -70,7 +83,7 @@ def _get_from_s3(self, file_prefix, file_name, file_type, tf):
 
     def _get_from_local(self, file_prefix, file_name, file_type):
         file_path = os.path.join(file_prefix, file_name)
-        file_content = open(file_path, 'rb').read()
+        file_content = open(file_path, 'rb')
```
hblanks: Do we ever close this file?

SamDepardieu: Good point

requirements.txt
```diff
@@ -8,7 +8,7 @@ psycopg2-binary==2.7.5
 python-dateutil==2.7.3
 pytz==2018.5
 s3transfer==0.1.13
-scikit-learn==0.20.0
+scikit-learn==0.19.2
```
nsorros: why this downgrade?

utils/serialiser.py
```diff
@@ -11,10 +11,10 @@ def serialise_matched_reference(data, current_timestamp):
 
 def serialise_reference(data, current_timestamp):
     """Serialise the data parsed by the model."""
-    if data.get('Title'):
+    if data.get('Title', '') and len(data.get('Title', '')) > 1024:
         title = data['Title'][:1024]
     else:
-        title = None
+        title = data.get('Title', None)
```
nsorros: this is a bit inconsistent with the above, in that here if there is not title we add a null whereas above we add an empty string

requirements.txt
SamDepardieu: The models have been trained with the version 0.19.2, using them on version 0.20.0 led to multiple warnings

nsorros: 👍

PR #24, opened 2018-12-10
Add the sentry module for exception reporting

SamDepardieu:
# Description

❗️ This shouldn't be merged before we the sentry dsn parameter is added as an env variable to the fargate relevant tasks. ❗️ 

Sentry will help us know when our projects are failing. This PR aims to add Sentry exception reporting to the reference parser.

## Type of change

- [x] :sparkles: New feature

# How Has This Been Tested?

Local test run

# Checklist:


main.py
```diff
@@ -13,9 +14,11 @@
...
 
 
+@report_exception
 def run_predict(scraper_file, references_file,
                 model_file, vectorizer_file):
     logger = settings.logger
+
     logger.setLevel('INFO')
     mode = 'S3' if settings.S3 else 'LOCAL'
     fm = FileManager(mode)
```
hblanks: If the main block is your only entry point, you may want to catch everything in the main block actually, instead of using a decorator.

```python
if __name__ == '__main__':
    try:
        ...
    except:
        caputure_exception()
        raise
```

sentry.py
```diff
@@ -0,0 +1,31 @@
...
+        except:
+            sentry_sdk.capture_exception()
+            raise
+
+    return wrapped_f
+
+
+# SENTRY_DSN must be present at import time. If we don't have it then,
+# we won't have it later either.
+init_sentry_sdk(os.environ['SENTRY_DSN'])
```
hblanks: This may as well be called at the start of the main block, actually, since it's less of a surprise there. We just didn't have that luxury with Airflow.

main.py
```diff
@@ -109,37 +111,45 @@ def run_predict(scraper_file, references_file,
...
+                    'run_predict(scraper_file, references_file,',
+                    'model_file, vectorizer_file)'
+                ]),
+                'stats_dumps'
+            )
+        else:
+            run_predict(scraper_file, references_file,
+                        model_file, vectorizer_file)
+    except Exception as e:
+        sentry_sdk.capture_exception(e)
```
nsorros: I assume in the future we need to capture exceptions where they happen right?

PR #28, opened 2019-01-16
Remove pipenv from this repository

SamDepardieu:
Pipenv made things more complicated for us in the end. This PR aims to get rid of it.
 - 🚚 Remove pipenv related files (`Pipfile` & `Pipfile.lock`)
 - 📝 Update related documentation
 - ✨ Add a `virtualenv` argument to the makefile to auto-generate the virtualenv


Makefile
```diff
@@ -1,5 +1,7 @@
 .DEFAULT_GOAL := all
 
+VIRTUALENV := env
```
hblanks: Might be good to follow the convention of datalabs/ repo and have this be `build/virtualenv`.

README.md
```diff
@@ -8,14 +8,15 @@ It also requires a PostgreSQL server to store the results on production.
...
+1. Python 3.5 or higher and `virtualenv`
 2. PostgreSQL 9 or higher
 3. A clean json file containing reference sections
 4. A clean csv file containing all your references
 
 Once you have everything installed, run:
-  * `pipenv install`
-  * `pipenv shell`
+  * `virtualenv env --python python3.6`
+  * `pip install -r requirements.txt`
```
hblanks: `make virtualenv` probably best here instead of separate commands.

Makefile
```diff
@@ -19,5 +21,15 @@ push: image
 	docker push $(ECR_IMAGE):$(VERSION) && \
 	docker push $(ECR_IMAGE):latest
 
+$(VIRTUALENV)/.installed: requirements.txt
+	@if [ -d $(VIRTUALENV) ]; then rm -rf $(VIRTUALENV); fi
+	@mkdir -p $(VIRTUALENV)
+	virtualenv --python python3.6 $(VIRTUALENV)
```
hblanks: not sure, but may need this to be `--python python3`

hblanks: (for better cross-platform support)

nsorros: plus if we tied to 3.6 we would run into the same problem we tried to solve which was that if someone has python 3.7 and not 3.6 it becomes an interesting exercise how to install python 3.6 just for the tool.

SamDepardieu: Fixed 👍

SamDepardieu:
🎳 💥 🤘

PR #36, opened 2019-01-22
Allowing the S3/local distinction to be specified in arguments of main

lizgzil:
# Description

This PR adapts mostly `main.py` so that it can be run with new arguments, these are:
- '--scraper-file'
- '--references-file'
- '--model-file'
- '--vectorizer-file'
- '--output-url'

Because S3 files start with 's3://' and the output url will start with 'file://' we can now distingush LOCAL and S3 runs of the code with these arguments rather than previously where we used variables in settings.

This means some parts of `settings.py` have been updated, and further work will also be needed.

I have also added `check_references_file` to `main.py` to make sure the user inputted references have the correct column names, and `get_file` to functionise a repeated process.

A small neatening of `models.py` is included in this PR.

## Type of change

- [x] :sparkles: New feature

# How Has This Been Tested?

I've run the code successfully with a mixture of S3 and local arguments.

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,21 @@ Once you're happy with your configuration, just run `python main.py`
 
 ### Method 3.
 
-Edit the `main.py` arguments in `run_parser.sh` to include the names of the input and output file locations desired. If the `scraper_file`, `references_file`, `model_file`, and `vectorizer_file` arguments are to S3 locations then make sure these start with `s3://`, otherwise file names are assumed to be locally stored. If the `output_url` argument is to a local location, then make sure it begins with `file://`, otherwise it is assumed to be from a database.
+Make an output folder `output_folder_name` and run `main.py` with arguments of your file locations, e.g. in the terminal run:
 
-Then run in the terminal `./run_parser.sh`. 
+```
+mkdir -p /tmp/parser-output/output_folder_name
+
+python3 $DIR/main.py \
+	--scraper-file "s3://datalabs-data/scraper-results/msf/20190117.json" \
+	--references-file "match-references/MRC_Publications_Nov2018_JGHT_JHSRI.csv" \
+	--model-file "s3://datalabs-data/reference_parser_models/RefSorter_classifier.pkl" \
+	--vectorizer-file "s3://datalabs-data/reference_parser_models/RefSorter_vectorizer.pkl" \
+	--output-url "file:///tmp/parser-output/output_folder_name"
+```
+
+If the `scraper_file`, `references_file`, `model_file`, and `vectorizer_file` arguments are to S3 locations then make sure these start with `s3://`, otherwise file names are assumed to be locally stored. If the `output_url` argument is to a local location, then make sure it begins with `file://`, otherwise it is assumed to be from a database.
+
 
 ## Contributing
 See the [Contributing guidelines](./CONTRIBUTING.md)
--- a/README.md
+++ b/README.md
@@ -43,5 +43,11 @@ This repository includes a `settings.py` file, where you can manually configure 
 
 Once you're happy with your configuration, just run `python main.py`
 
+### Method 3.
+
+Edit the `main.py` arguments in `run_parser.sh` to include the names of the input and output file locations desired. If the `scraper_file`, `references_file`, `model_file`, and `vectorizer_file` arguments are to S3 locations then make sure these start with `s3://`, otherwise file names are assumed to be locally stored. If the `output_url` argument is to a local location, then make sure it begins with `file://`, otherwise it is assumed to be from a database.
+
+Then run in the terminal `./run_parser.sh`. 
+
 ## Contributing
 See the [Contributing guidelines](./CONTRIBUTING.md)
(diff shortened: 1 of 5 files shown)


main.py
```diff
@@ -15,37 +17,55 @@
...
+    assert 'uber_id' in ref_file, (
+        "ref_file.uber_id not defined in " +
+        references_file +
+        " consider renaming current uber id column name to 'uber_id'"
+    )
 
-    logger.setLevel('INFO')
-    mode = 'S3' if settings.S3 else 'LOCAL'
-    fm = FileManager(mode)
+def get_file(file_str, file_type, get_scraped = None):
```
hblanks: Three thoughts here:

- If you keep it, `get_scraped=False` instead of `get_scraped = None`
- but, may be better to return (`fm`, `file_name`, `file_dir`) and let the calling function call `get_file()` or `get_scraping_results()`. (Just that keyword args are usually an issue)
- `file` is a builtin, so would be best not to use it also as a local variable. So, if you keep it, `result = fm.get_file(...)`, etc.

main.py
```diff
@@ -15,37 +17,55 @@
...
 
     # Loading the scraper results
-    scraper_file_name = os.path.basename(scraper_file)
-    scraper_file_dir = os.path.dirname(scraper_file)
-
-    scraper_file = fm.get_scraping_results(
-        scraper_file_name,
-        scraper_file_dir,
-    )
+    scraper_file = get_file(scraper_file, "", True)
```
hblanks: If you keep the keyword arg, `get_file(scraper_file, "", get_scraped=True)` would be more explicit.

If you remove it, this would become:

```python
fm, file_name, file_dir = get_file_manager(scraper_file, "")
scarper_file = fm.get_scraping_results(file_name, file_dir)
```

Either way's OK by me.

README.md
```diff
@@ -43,5 +43,23 @@ This repository includes a `settings.py` file, where you can manually configure
 
 Once you're happy with your configuration, just run `python main.py`
 
+### Method 3.
```
nsorros: Is the method 2 still relevant? Can someone run the code just by tweaking settings?

main.py
```diff
@@ -15,37 +17,55 @@
 from settings import settings
 
 
-def run_predict(scraper_file, references_file,
-                model_file, vectorizer_file):
-    logger = settings.logger
+def check_references_file(ref_file, references_file):
```
nsorros: is that check part of this PR?

settings.py
```diff
@@ -17,17 +17,17 @@ class BaseSettings:
...
-    REFERENCES_DIR = "wellcome_publications"
+    REFERENCES_DIR = "s3://{}/wellcome_publications".format(BUCKET)
     REFERENCES_FILENAME = 'uber_api_publications.csv'
 
     LOCAL_OUTPUT_DIR = 'local_output'
     PREF_REFS_FILENAME = 'predicted_reference_structures.csv'
     MATCHES_FILENAME = 'all_match_data.csv'
 
-    MODEL_DIR = "reference_parser_models"
+    MODEL_DIR = "s3://{}/reference_parser_models".format(BUCKET)
```
nsorros: so now the default is to pull these from s3? i assume it gets the latest as the prod does. one minor concern with this is that the default user outside wellcome might not use s3.

main.py
```diff
@@ -111,45 +134,92 @@ def run_predict(scraper_file, references_file,
 
 
 if __name__ == '__main__':
+    logger = settings.logger
+    logger.setLevel('INFO')
+
     # SENTRY_DSN must be present at import time. If we don't have it then,
     # we won't have it later either.
-    sentry_sdk.init(os.environ['SENTRY_DSN'])
+    if 'SENTRY_DSN' in os.environ:
```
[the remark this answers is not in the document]
hblanks: Yes

settings.py
hblanks: Or these models should be in a public S3 bucket, but, yes.

For now, the default is as it runs in production. This was not a change; it's just more explicit now than it was.

main.py
lizgzil: yes it is

README.md
lizgzil: You can run `main.py` with no arguments, so we could keep method 2 in, but maybe this is a duplication? I had to change things in settings to get it to run though. Happy to delete if you think I should?

main.py
lizgzil: i've chosen to keep it and make your suggested changes

main.py
lizgzil: i've done the former

main.py
```diff
@@ -29,7 +29,7 @@ def check_references_file(ref_file, references_file):
         " consider renaming current uber id column name to 'uber_id'"
     )
 
-def get_file(file_str, file_type, get_scraped = None):
+def get_file(file_str, file_type, get_scraped = False):
```
hblanks: a strange pythonism is that we normally don't put spaces in between `=` and keyword arguments (but we do for assignment statements, of course). But, OK to merge and correct this some other time. Thanks for touching itup.

PR #38, opened 2019-01-25
Refactor predict refs

lizgzil:
# Description

This PR:

- Rewrites `predict_references` so that it doesn't use pandas
- Rewrites `predict_references` so that it inputs a list of components and outputs a list of tuples (prediction, probability)
- Introduces some tests for `predict_references`
- Adds a line at the beginning of `predict_structure` so that the new input (which is now a dictionary) is converted to a pd dataframe, so that this function continues to work. The aim is to eventually (in a different PR) refactor `predict_structure` so that it doesn't require a pd dataframe input, so eventually this will be redundant.
- `main` had to be updated in order to get the new output of `predict_references` into the correct format for the rest of the code.
- Fixes a small error in the README

# How Has This Been Tested?

`test_predict.py` file has been added in the tests folder. You can run those tests by running `python -m unittest tests.test_predict` while on the root directory.

# Checklist:

### DIFF
--- a/main.py
+++ b/main.py
@@ -85,9 +85,13 @@ def run_predict(scraper_file, references_file,
     )
     
     # Link predictions back with all original data (Document id etc)
+    # When we merge and splited_references is a dict not a dataframe then we could just merge the list of dicts
     reference_components_predictions = splited_references
-    reference_components_predictions["Predicted Category"] = [tup[0] for tup in components_predictions]
-    reference_components_predictions["Prediction Probability"] = [tup[1] for tup in components_predictions]
+    reference_components_predictions["Predicted Category"] = (
+        [components_prediction["Predicted Category"] for components_prediction in components_predictions]
+        )
+    reference_components_predictions["Prediction Probability"] = (
+        [components_prediction["Prediction Probability"] for components_prediction in components_predictions])
     
     # Predict the reference structure????
     predicted_reference_structures = predict_structure(
--- a/main.py
+++ b/main.py
@@ -121,7 +121,7 @@ def run_predict(scraper_file, references_file,
         splitted_components = process_references(splitted_references)
 
         # Predict the references types (eg title/author...)
-        logger.info('[+] Predicting the reference components')
+        # logger.info('[+] Predicting the reference components')
         components_predictions = predict_references(
             pool_map,
             mnb,
--- a/main.py
+++ b/main.py
@@ -78,12 +78,17 @@ def run_predict(scraper_file, references_file,
 
     # Predict the references types (eg title/author...)
     logger.info('[+] Predicting the reference components')
-    reference_components_predictions = predict_references(
+    components_predictions = predict_references(
         mnb,
         vectorizer,
-        splited_references
+        splited_references['Reference component']
     )
-
+    
+    # Link predictions back with all original data (Document id etc)
+    reference_components_predictions = splited_references
+    reference_components_predictions["Predicted Category"] = [tup[0] for tup in components_predictions]
+    reference_components_predictions["Prediction Probability"] = [tup[1] for tup in components_predictions]
+    
     # Predict the reference structure????
     predicted_reference_structures = predict_structure(
         reference_components_predictions,
--- a/main.py
+++ b/main.py
@@ -87,11 +87,12 @@ def run_predict(scraper_file, references_file,
     # Link predictions back with all original data (Document id etc)
     # When we merge and splited_references is a dict not a dataframe then we could just merge the list of dicts
     reference_components_predictions = splited_references
-    reference_components_predictions["Predicted Category"] = (
-        [components_prediction["Predicted Category"] for components_prediction in components_predictions]
-        )
-    reference_components_predictions["Prediction Probability"] = (
-        [components_prediction["Prediction Probability"] for components_prediction in components_predictions])
+    reference_components_predictions["Predicted Category"] = [
+        d["Predicted Category"] for d in components_predictions
+    ]
+    reference_components_predictions["Prediction Probability"] = [
+        d["Prediction Probability"] for d in components_predictions
+    ]
     
     # Predict the reference structure????
     predicted_reference_structures = predict_structure(
--- a/tests/test_predict.py
+++ b/tests/test_predict.py
@@ -21,7 +21,10 @@ class TestPredict(unittest.TestCase):
 		components_predictions = predict_references(mnb,vectorizer,["Component","Component","Component"])
 		all_categories = ['Authors', 'Title', 'Journal', 'PubYear', 'Volume', 'Issue', 'Pagination']
 		self.assertEqual(
-			isinstance(components_predictions, list) and
+			isinstance(components_predictions, list),
+			True, "Should be a list"
+			)
+		self.assertEqual(
 			all([components_prediction[0] in all_categories for components_prediction in components_predictions]),
 			True, "Should be a list of categories from {}".format(all_categories)
 			)
--- a/tests/test_predict.py
+++ b/tests/test_predict.py
@@ -4,7 +4,7 @@ import pickle
 
 from utils import predict_references
 
-MODELS_DIR = os.path.join( os.path.dirname(__file__), '..', 'reference_parser_models')
+MODELS_DIR = os.path.join(os.path.dirname(__file__), '..', 'reference_parser_models')
 
 with open(os.path.join(MODELS_DIR, "RefSorter_classifier.pkl"), "rb") as f:
     mnb = pickle.load(f)
--- a/tests/test_predict.py
+++ b/tests/test_predict.py
@@ -14,31 +14,31 @@ with open(os.path.join(MODELS_DIR, "RefSorter_vectorizer.pkl"), "rb") as f:
 
 class TestPredict(unittest.TestCase):
 	def test_empty_components(self):
-		components_predictions = predict_references(mnb,vectorizer,[])
+		components_predictions = predict_references(map,mnb,vectorizer,[])
 		self.assertEqual(components_predictions, [], "Should be []")
 
 	def test_normal_components(self):
-		components_predictions = predict_references(mnb,vectorizer,["Component","Component","Component"])
+		components_predictions = predict_references(map,mnb,vectorizer,["Component","Component","Component"])
 		all_categories = ['Authors', 'Title', 'Journal', 'PubYear', 'Volume', 'Issue', 'Pagination']
 		self.assertEqual(
 			isinstance(components_predictions, list),
 			True, "Should be a list"
 			)
 		self.assertEqual(
-			all([components_prediction[0] in all_categories for components_prediction in components_predictions]),
+			all([components_prediction['Predicted Category'] in all_categories for components_prediction in components_predictions]),
 			True, "Should be a list of categories from {}".format(all_categories)
 			)
 
 	def test_size(self):
-		components_predictions = predict_references(mnb,vectorizer,["Component","Component","Component"])
+		components_predictions = predict_references(map,mnb,vectorizer,["Component","Component","Component"])
 		self.assertEqual(len(components_predictions) , 3, "Should be 3")
 
 	def test_string_component(self):
-		components_predictions = predict_references(mnb,vectorizer,["Component"])
-		self.assertEqual(isinstance(components_predictions[0][0], str), True, "Should be a string")
+		components_predictions = predict_references(map,mnb,vectorizer,["Component"])
+		self.assertEqual(isinstance(components_predictions[0]['Predicted Category'], str), True, "Should be a string")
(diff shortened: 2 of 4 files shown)


tests/test_predict.py
```diff
@@ -0,0 +1,32 @@
...
+
+from utils import predict_references
+
+import pickle
+
+pickle_in = open("./reference_parser_models/RefSorter_classifier.pkl","rb")
+mnb = pickle.load(pickle_in)
+
+pickle_in = open("./reference_parser_models/RefSorter_vectorizer.pkl","rb")
+vectorizer = pickle.load(pickle_in)
```
hblanks: Might not be the worst thing to make this code path independent. Also, best is to close files when you're done with them.

```python
import os.path
import pickle
import unittest

from utils import predict_references

MODELS_DIR = os.path.join( os.path.dirname(__file__), '..', 'reference_parser_models')

with open(os.path.join(MODELS_DIR, "RefSorter_classifier.pkl"), "rb") as f:
    mnb = pickle.load(f)

...
```

utils/predict.py
```diff
@@ -266,20 +266,17 @@ def predict_references(mnb,
...
 
-    predict_all = pd.DataFrame.from_dict(predict_all)
-
-    reference_components_predictions = reference_components.reset_index()
+    components_predictions = []
+    for dic in predict_all:
+        tup = ()
+        for key,val in dic.items():
+            tup = (*tup,val)
+        components_predictions.append(tup)
```
hblanks: I'm not sure I quite get what's going on here. Maybe there's a cleaner sort of list comprehension or itertools function we could be using? Or, we could just append things to a list and convert to a tuple at the end?

tests/test_predict.py
```diff
@@ -0,0 +1,32 @@
...
+
+class TestSplit(unittest.TestCase):
+	def test_empty_components(self):
+		components_predictions = predict_references(mnb,vectorizer,"")
```
nsorros: should'nt this be an empty list instead of an empty string?

tests/test_predict.py
```diff
@@ -0,0 +1,32 @@
...
+		self.assertEqual(components_predictions, [], "Should be []")
+
+	def test_normal_components(self):
+		components_predictions = predict_references(mnb,vectorizer,["Component","Component","Component"])
+		self.assertEqual(isinstance(components_predictions, list) , True, "Should be a list")
```
nsorros: here actually instead of testing whether it returns a list only which you should you can also check whether the predictions are from a list of categories, for example tha are either author, title etc

tests/test_predict.py
```diff
@@ -0,0 +1,32 @@
...
+
+	def test_size(self):
+		components_predictions = predict_references(mnb,vectorizer,["Component","Component","Component"])
+		self.assertEqual(len(components_predictions) , 3, "Should be 3")
+
+	def test_author_component(self):
```
nsorros: how is that testing the author component?

tests/test_predict.py
```diff
@@ -0,0 +1,32 @@
...
+
+import pickle
+
+pickle_in = open("./reference_parser_models/RefSorter_classifier.pkl","rb")
+mnb = pickle.load(pickle_in)
+
+pickle_in = open("./reference_parser_models/RefSorter_vectorizer.pkl","rb")
+vectorizer = pickle.load(pickle_in)
+
+class TestSplit(unittest.TestCase):
```
nsorros: This should probably be named TestPredict, be careful when copying and pasting.

utils/predict.py
```diff
@@ -240,14 +241,13 @@ def _get_year_or_component(component, mnb, vectorizer):
 
 def predict_references(mnb,
                        vectorizer,
-                       reference_components,
+                       reference_components_list,
```
nsorros: maybe list is not needed in the name. add a docstring where you describe the input and the output which would be better

utils/predict.py
```diff
@@ -266,20 +266,17 @@ def predict_references(mnb,
         partial(_get_year_or_component,
```
nsorros: maybe predict reference compnent instead of get year or component

utils/predict.py
```diff
@@ -146,6 +144,9 @@ def predict_structure(reference_components_predictions,
     each reference for each document in turn.
     """
 
+    # Convert to pd dataframe, although when this function is refactored we shouldnt have to do this
+    reference_components_predictions = pd.DataFrame.from_dict(reference_components_predictions)
+
```
nsorros: In regards to simplify the prediction part

This can probably be in a function is_year
```
if (
       (component.isdecimal()
        and int(component) in valid_years_range)
       or (
           len(component) == 6
           and component[1:5].isdecimal()
           and int(component[1:5]) in valid_years_range)
       ):
```

nsorros: Then maybe in the if then else write two variables pred_cat and pred_prob and always return the structure in the end like
```
return {
            'Predicted Category': pred_cat,
            'Prediction Probability': pred_prob
        }

Comment (nsorros) [reply utils/predict.py]: The main thing I am trying to do is to make it a bit more readable.

File: tests/test_predict.py (hunk 1 above)
Comment (lizgzil) [reply tests/test_predict.py]: 👍 Is there a reason for the space before ` os.path.dirnam`?

File: tests/test_predict.py (hunk 3 above)
Comment (lizgzil) [reply tests/test_predict.py]: yes that is true

File: tests/test_predict.py (hunk 4 above)
Comment (lizgzil) [reply tests/test_predict.py]: Is this what you mean:
```
all_categories = ['Authors', 'Title', 'Journal', 'PubYear', 'Volume', 'Issue', 'Pagination']
self.assertEqual(
        isinstance(components_predictions, list) and
	all([c[0] in all_categories for c in components_predictions]),
	True, "Should be a list of categories from {}".format(all_categories)
        )
```
Although is this test not a bit risky if out list of `all_categories` or spellings of the categories change when we build the model?

File: tests/test_predict.py (hunk 5 above)
Comment (lizgzil) [reply tests/test_predict.py]: 👍 oops, this is from a previous test idea. I've changed this to `test_string_component`

File: utils/predict.py (hunk 7 above)
Comment (lizgzil) [reply utils/predict.py]: 👍

File: tests/test_predict.py (hunk 4 above)
Comment (nsorros) [reply tests/test_predict.py]: Yes, but i think it would be better to add the second check in a second assertEqual instead.

If we change how name the categories we would need to change the test, that is ok.

File: utils/predict.py (hunk 2 above)
Comment (lizgzil) [reply utils/predict.py]: The point was to convert a list of dicts to a list of tuples. I googled, but couldnt find a nicer way to do this. I've added this to a new function `dict_to_tuples`, so perhaps this will make it clearer that this is a generic function. @nsorros any ideas of how to make this function nicer?

File: tests/test_predict.py (hunk 4 above)
Comment (lizgzil) [reply tests/test_predict.py]: :)

**Comment by lizgzil** [comment]
@nsorros @hblanks thanks :) I've address your comments, please can you see what you think

File: tests/test_predict.py (hunk 1 above)
Comment (hblanks) [reply tests/test_predict.py]: Sorry, no - just an oversight on my part. Editing now.

File: utils/predict.py (hunk 2 above)
Comment (hblanks) [reply utils/predict.py]: Maybe `component_predictions = [ tuple(val) for val in dic.values() ]` ?

File: utils/predict.py [hunk 10]
```diff
@@ -206,48 +207,77 @@ def predict_reference_comp(mnb, vectorizer, word_list):
...
+    """
+
+    list_tuples = []
+    for dic in list_dict:
+        tup = ()
+        for key,val in dic.items():
+            tup = (*tup,val)
+        list_tuples.append(tup)
+
+    return list_tuples
```
hblanks: Noted earlier, I think this is a list comprehension...slightly different than last comment though:

`return [tuple(d.values()) for d in list_dict]`

lizgzil: lovely, thanks :)

main.py
```diff
@@ -78,12 +78,17 @@ def run_predict(scraper_file, references_file,
...
         vectorizer,
-        splited_references
+        splited_references['Reference component']
     )
-
+    
+    # Link predictions back with all original data (Document id etc)
+    reference_components_predictions = splited_references
+    reference_components_predictions["Predicted Category"] = [tup[0] for tup in components_predictions]
+    reference_components_predictions["Prediction Probability"] = [tup[1] for tup in components_predictions]
```
nsorros: last thing, you do not need to construct this here, create it inside predict_references. instead of returning a list of tuples return a list of dicts with the names as written here.

lizgzil: ok, so no more tuples

nsorros: not yet since there is little point at the moment to go from tuples to dicts in the end.

lizgzil: i've addressed this but not sure if it's exactly what youd hoped. I think when `splited_references` is a dict not a pd df (as it was when I made this branch) then the extra lines in `main` could be made nicer since I think merging 2 lists of dicts is easier than 1 pd df and 1 list of dicts.

main.py
```diff
@@ -85,9 +85,13 @@ def run_predict(scraper_file, references_file,
     )
     
     # Link predictions back with all original data (Document id etc)
+    # When we merge and splited_references is a dict not a dataframe then we could just merge the list of dicts
     reference_components_predictions = splited_references
-    reference_components_predictions["Predicted Category"] = [tup[0] for tup in components_predictions]
-    reference_components_predictions["Prediction Probability"] = [tup[1] for tup in components_predictions]
+    reference_components_predictions["Predicted Category"] = (
```
nsorros: do you need to do that?

lizgzil: I thought that was the way to get all the values for one key out of a list of dicts. e.g. https://stackoverflow.com/questions/7271482/python-getting-a-list-of-value-from-list-of-dict

lizgzil: 👍 thanks. Is there a rule for when you can use shorter names like this? Or is it just aesthetic choice for when variable names are long?

lizgzil:
@hblanks and @nsorros thanks for the reviews. I've addressed your comments thus far in my latest commits.

main.py
hblanks: purely aesthetic choice, and it's somewhat language dependent, too.

Python tends to favor meaningful variable names more than short ones, with an exception like this being often made for list comprehensions / generator expressions, because the name is obvious and the line's shorter with a short name.

As a point of contrast: Go tends to use single-character variable names much more often, and often as an idiom. So, almost all `*http.Request` variable in go are just named `r`. Example:

```go
http.HandleFunc("/bar", func(w http.ResponseWriter, r *http.Request) {
	fmt.Fprintf(w, "Hello, %q", html.EscapeString(r.URL.Path))
})
```
(https://golang.org/pkg/net/http/)

main.py
```diff
@@ -111,20 +111,35 @@ def run_predict(scraper_file, references_file,
...
             settings.ORGANISATION_REGEX
         )
 
-        # logger.info('[+] Parsing the references')
+        # Split references into components
         splitted_components = process_references(splitted_references)
-        reference_components_predictions = predict_references(
+
+        # Predict the references types (eg title/author...)
+        logger.info('[+] Predicting the reference components')
```
nsorros: Since we do not log for every document now
# logger.info('[+] Parsing the references')

I think you should comment this
logger.info('[+] Predicting the reference components')

utils/predict.py
```diff
@@ -241,70 +242,73 @@ def predict_reference_comp(mnb, vectorizer, word_list):
...
                        vectorizer,
                        reference_components):
 
+    """
+    Predicts the categories for a list of reference components.
+    Input:
+    - mnb: The trained multinomial naive Bayes model for predicting the categories of reference components
+    - vectorizer: The vector of word counts in the training set
+    - reference_components: A list of reference components
+    - num_workers: How many different processors you want to use in multiprocessing the predicting
```
nsorros: this is out of date. update it with the latest arg

hblanks:
pretty sure we're gonna have a bad merge conflict with https://github.com/wellcometrust/policytool/pull/40 

but, that's a good lesson for us all about the WIP, and I'll be happy to help with it.

Merge conflicts like this really benefit from dedicated 3-way viewer. So, before you work on it, you may want to [download P4merge](https://www.perforce.com/downloads/visual-merge-tool) and add something like this to your `.gitconfig` beforehand:

```
[merge]
	tool = p4mergetool
[mergetool "p4mergetool"]
	cmd = /opt/perforce/p4v/bin/p4merge $PWD/$BASE $PWD/$REMOTE $PWD/$LOCAL $PWD/$MERGED
	trustExitCode = false
[mergetool]
	keepBackup = false
```

main.py
lizgzil: 👍

utils/predict.py
lizgzil: 👍

PR #42, opened 2019-01-30
Fix #41 to deal with case of Title prediction

### LINKED ISSUE #41: Fix main to account for times when no Title column is predicted
Now that we process each policy document one by one we run into a problem occasionally where no Title is predicted in the parsing of references, and therefore the fuzzy matching step will break (since there is no 'Title' column to match on).
e.g.
screenshot 2019-01-30 at 15 32 27

If this occurs the fuzzy match step should be skipped in main and a blank `all_match_data.csv` should be created (or not created at all).

lizgzil:
Fix for https://github.com/wellcometrust/reference-parser/issues/41

# Description

Small change to `predict.py` and `settings.py` to make output contain all categories even if nothing is predicted for any of them. As in the case where no matched references are found, a file called `all_match_data.csv` is still outputted, but it will be empty.

This looks like more changes than there were due to a tab change.

## Type of change

# How Has This Been Tested?

It hasn't.

# Checklist:

### DIFF
--- a/utils/predict.py
+++ b/utils/predict.py
@@ -71,10 +71,11 @@ def decide_components(single_reference):
     single_reference = pd.concat([single_reference, block_number], axis=1)
     single_reference_components = {}
 
-    for classes in set(single_reference["Predicted Category"]):
+    categories =  settings.REF_CLASSES
+    for category in categories:
 
         # Are there any sentences of this type (i.e. Authors, Title)?
-        classornot = sum(single_reference['Predicted Category'] == classes)
+        classornot = sum(single_reference['Predicted Category'] == category)
 
         if classornot != 0:
 
@@ -82,18 +83,18 @@ def decide_components(single_reference):
             number_blocks = len(
                 single_reference['Block'][single_reference[
                     'Predicted Category'
-                ] == classes].unique()
+                ] == category].unique()
             )
 
             if number_blocks == 1:
                 # Just record this block separated by commas
                 single_reference_components.update({
-                    classes: ", ".join(
+                    category: ", ".join(
                         single_reference[
                             'Reference component'
                         ][single_reference[
                             'Predicted Category'
-                        ] == classes]
+                        ] == category]
                     )
                 })
             else:
@@ -104,7 +105,7 @@ def decide_components(single_reference):
                 # (random choice)
                 
                 highest_probability_index = single_reference[
-                    single_reference['Predicted Category'] == classes
+                    single_reference['Predicted Category'] == category
                 ]['Prediction Probability'].idxmax()
 
                 highest_probability_block = single_reference[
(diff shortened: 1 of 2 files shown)


utils/fuzzymatch.py
```diff
@@ -57,27 +57,29 @@ def fuzzy_match_blocks(self, blocksize, predicted_publications, threshold):
...
-        nextblocksize = blocksize
-
         all_match_data = []
-        while counter < len(predicted_publications):
-            match_data = self.match_vectorised(
-                predicted_publications[counter:(counter + nextblocksize)],
-                threshold
-            )
+        
+        if 'Title' in predicted_publications:
```
nsorros: my only comment is that this function does not use the title column so this check should probably go to match_vectorised and in the beggining where you check if there is a title and if not you return no match

lizgzil: 👍

utils/fuzzymatch.py
```diff
@@ -18,37 +18,41 @@ def __init__(self, real_publications, titles):
...
-            predicted_publications['Title']
-        )].reset_index()
 
-        title_vectors = self.vectorizer.transform(
-            predicted_publications['Title']
-        )
-        title_similarities = cosine_similarity(
-            title_vectors, self.tfidf_matrix
-        )
+        if 'Title' in predicted_publications:
```
nsorros: I am slightly inclined to suggest that the predictions should always return all categories. What do you think? Otherwise, we would need to fill the code with a couple of ifs whenever we need one category or the other.

I always thought that we should guarantee the categories are present but happy to discuss whether this makes sense.

hblanks:
@lizgzil , @nsorros  - just checking in since this PR's been out 2 weeks. Anything we can be doing to help it merge?

lizgzil:
> @lizgzil , @nsorros - just checking in since this PR's been out 2 weeks. Anything we can be doing to help it merge?

I will speak to @nsorros about this tomorrow

utils/predict.py
```diff
@@ -71,29 +71,30 @@ def decide_components(single_reference):
     single_reference = pd.concat([single_reference, block_number], axis=1)
     single_reference_components = {}
 
-    for classes in set(single_reference["Predicted Category"]):
+    categories =  settings.REF_CLASSES
+    for category in categories:
 
         # Are there any sentences of this type (i.e. Authors, Title)?
-        classornot = sum(single_reference['Predicted Category'] == classes)
+        classornot = sum(single_reference['Predicted Category'] == category)
```
nsorros: I don't understand how a change in a variable name fixes the problem

lizgzil: lines 74 and 75 are the changes that fix the problem, but when I was doing it I decided to change the variable name too

PR #50, opened 2019-02-06
Rename variables, functions and file in separate

nsorros:
This PR zooms into the separate file and inspects the variables name to make sure they are easily understandable to a new user and are in line with how we communicate about the product externally.

Please do leave comments related to the changes I made and also if you spot anything that you be improved in terms of naming. If something is not clear now, it will probably not be in the future either so mention it.

I chose the name split for the name instead of separate since this is how we communicate this module in our presentations.

### DIFF
--- a/main.py
+++ b/main.py
@@ -13,7 +13,7 @@ import sentry_sdk
 
 from utils import (FileManager,
                    FuzzyMatcher,
-                   process_reference_section,
+                   process_references_section,
                    process_references,
                    predict_references,
                    predict_structure)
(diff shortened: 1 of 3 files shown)


main.py
```diff
@@ -123,8 +123,9 @@ def run_predict(scraper_file, references_file,
             pool_map,
             mnb,
             vectorizer,
-            splitted_components
+            splitted_components['Reference component']
```
nsorros: this is possible an error from cheery picking the commit from somewhere else. will fix.

PR #54, opened 2019-02-07
Remove pipenv from the web scraper

SamDepardieu:
# Description

Replace pipenv back to a normal virtualenv.
Also include a Makefile command to setup the virtualenv.

## Type of change

- [x] :fire: Breaking change


wsf-scraper-web/Dockerfile
```diff
@@ -10,9 +10,7 @@ COPY ./entrypoint.sh /wsf_scraper/entrypoint.sh
 
 COPY ./scrapy.cfg /wsf_scraper/scrapy.cfg
 
-# Pipenv require these two files to install requirements in the pinned version
-COPY ./Pipfile /wsf_scraper/Pipfile
-COPY ./Pipfile.lock /wsf_scraper/Pipfile.lock
+COPY ./requirements.txt /wsf_scraper/requirements.txt
```
nsorros: why the duplication since you have a dockerfile.base?

SamDepardieu: Good question, I don't remember why we kept the two files 🤔 
@hblanks Should we delete this one?

wsf-scraper-web/Pipfile
```diff
@@ -1,19 +0,0 @@
...
-scrapy = "==1.5.1"
-"psycopg2-binary" = "==2.7.5"
-pyahocorasick = "==1.1.8"
-requests = "==2.20.0"
-"bs4" = "*"
-"boto3" = "==1.9.1"
-sentry-sdk = "==0.6.2"
-
-[dev-packages]
-pytest = "*"
```
nsorros: this is not in the requirements? so either add it in the documentation or in the requirements

wsf-scraper-web/Pipfile
```diff
@@ -1,19 +0,0 @@
...
-
-[requires]
-python_version = "3.6"
```
nsorros: did you update the documentation to say that you need 3.6 to run?

wsf-scraper-web/Pipfile
SamDepardieu: I'm not using pytest with the scraper, so I removed it from the requirements as well.

wsf-scraper-web/Pipfile
SamDepardieu: The documentation is focused on running it with docker compose, so the Python version is set in the dockerfile

hblanks: Not sure we need 3.6 to run...do we?

SamDepardieu: Any python3 is fine, I think

nsorros: Just document it is python 3 somewhere, i guess

PR #56, opened 2019-02-08
Merge match results from multiple documents and organisations

lizgzil:
# Description

This PR adds a python file to merge all the matched files for each documents and each organisation together into one. It joins the merged outputs with the urls and the references information. Most of the file names etc are user inputted.

I find that some of the _predicted_reference_structures files don't read in without errors (so I add a 'try'). This happens on 4 occasions and the error is:

screenshot 2019-02-13 at 16 19 00

## Type of change
- [ ] :sparkles: New feature

# How Has This Been Tested?

No testing has been done.

### DIFF
--- a/merge_results.py
+++ b/merge_results.py
@@ -4,6 +4,7 @@ Manually edit main with the folder names you want to merge all together
 Output is one csv of all the matches from all the organisations you have provided
 """
 
+from argparse import ArgumentParser
 import fnmatch
 import os
 import glob
@@ -27,7 +28,7 @@ def get_csv_names(file_dir, suffix):
 
     return csv_names
 
-def merge_match_csvs(match_csv_names):
+def concat_match_csvs(match_csv_names):
     """
     This function merges all the csvs given in match_csv_names
     Input:
@@ -41,12 +42,12 @@ def merge_match_csvs(match_csv_names):
         if not match_data.empty:
             all_matches.append(match_data)
 
-    all_matches = pd.concat(all_matches, ignore_index=False)
+    all_matches = pd.concat(all_matches)
 
     return all_matches
 
 
-def merge_predicted_csvs(predicted_csv_names):
+def concat_predicted_csvs(predicted_csv_names):
     """
     This function gets all the document id & document url from the predicted references csv
     Input:
@@ -70,21 +71,35 @@ def merge_predicted_csvs(predicted_csv_names):
 
     return all_url
 
+def create_argparser(description):
+    parser = ArgumentParser(description)
+    parser.add_argument(
+        '--file_dir',
+        help='Directory of saved output folders',
+        default = './tmp/parser-output/charlene'
+    )
+    parser.add_argument(
+        '--match_refs_file',
+        help='Original references file which was used in parsing of saved outputs',
+        default = './match-references/charlene_publications_format.csv'
+    )
+    return parser
+
 if __name__ == '__main__':
 
-    file_dir = './tmp/parser-output/charlene' # "pophealth", "charlene"
-    output_name = 'merged_all_matches'
+    parser = create_argparser(__doc__.strip())
+    args = parser.parse_args()
+
+    file_dir = args.file_dir
+    output_folder_name = 'merged_all_matches'
     
-    match_refs = pd.read_csv(
-        "./match-references/charlene_publications_format.csv")
-    # "./match-references/MRC_Publications_Nov2018_JGHT_JHSRI_all.csv"
-    # "./match-references/charlene_publications_format.csv"
+    match_refs = pd.read_csv(args.match_refs_file)
 
     match_csv_names = get_csv_names(file_dir, '_all_match_data')
     predicted_csv_names = get_csv_names(file_dir, '_predicted_reference_structures')
 
-    all_match = merge_match_csvs(match_csv_names)
-    all_url = merge_predicted_csvs(predicted_csv_names)
+    all_match = concat_match_csvs(match_csv_names)
+    all_url = concat_predicted_csvs(predicted_csv_names)
 
     all_matches_url = all_match.join(
         all_url.set_index('Document id'),
@@ -96,9 +111,9 @@ if __name__ == '__main__':
         on='WT_Ref_Id'
         ) # Join with references information
 
-    if not os.path.exists('{}/{}'.format(file_dir, output_name)):
-            os.makedirs('{}/{}'.format(file_dir, output_name))
+    if not os.path.exists('{}/{}'.format(file_dir, output_folder_name)):
+            os.makedirs('{}/{}'.format(file_dir, output_folder_name))
 
-    all_matches_refs.to_csv('{}/{}/merged_all_matches.csv'.format(file_dir, output_name))
+    all_matches_refs.to_csv('{}/{}/merged_all_matches.csv'.format(file_dir, output_folder_name))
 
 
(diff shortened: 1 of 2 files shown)


merge_results.py
```diff
@@ -0,0 +1,70 @@
...
+        "/Users/gallaghe/Code/reference-parser/match-references/charlene_publications_format.csv")
+    # "/Users/gallaghe/Code/reference-parser/match-references/MRC_Publications_Nov2018_JGHT_JHSRI_all.csv"
+
+    folder_name = "/Users/gallaghe/Code/reference-parser/tmp/parser-output"
+    who_iris = "{}_who_iris20190207".format(job_name)
+    unicef = "{}_unicef20190131".format(job_name)
+    msf = "{}_msf20190131".format(job_name)
+    nice = "{}_nice20190131".format(job_name)
+
+    all_matches_refs = merge_matches(folder_name, [who_iris, unicef, msf, nice], match_refs)
```
nsorros: have you tried doing this by using cat like this `cat *.csv > all_matches_charlene.csv` or something?

merge_results.py
```diff
@@ -0,0 +1,70 @@
...
+                all_matches.append(match_data)
+                pred_data = pd.read_csv("{}/{}_predicted_reference_structures.csv".format(file_dir,file.split("_")[0]))
+                all_uri.append(pred_data[['Document id','Document uri']].drop_duplicates())
+
+    return all_matches, all_uri
+
+def merge_matches(folder_name, org_list, match_refs):
+    """
+    Get all the _all_match_data.csv files for all organisations
+    Concat them all and join to the original references data
```
nsorros: why do you join with the original reference data?

lizgzil: Without joining you only get the uber_id and the title. Thought it'd be useful for Charlene and Elena to be able to see the journals, author lists, or whatever else their original data had.

merge_results.py
lizgzil: I haven't, but I don't want to concat any empty csvs, and many of the `..._all_matches.csv` files are empty. Would this method add many blank rows, or just ignore?

Struggling to find a way to both match a folder name and a file within that folder :(, since that would at least takeaway the need to list the folder names.

lizgzil: @nsorros  i've made this a little neater in the latest commit

merge_results.py
```diff
@@ -0,0 +1,70 @@
...
+    return all_matches_refs
+
+if __name__ == '__main__':
+
+    # The name you used in your output file name
+    job_name = "charlene" # "pophealth", "charlene"
+    
+    # The original references data searched for
+    match_refs = pd.read_csv(
+        "/Users/gallaghe/Code/reference-parser/match-references/charlene_publications_format.csv")
```
nsorros: this is a duplicate with a comment bellow. i think the comments needs to go and possibly you need to add an argument for this, if you want other people to use it

merge_results.py
```diff
@@ -0,0 +1,70 @@
...
+    Concat them all and join to the original references data
+    """
+    all_matches = []
+    all_uri = []
+    for folder_name in fnmatch.filter(os.listdir(folder_dir), '{}_*'.format(job_name)):
+        if os.path.isdir(folder_dir+folder_name):
+            file_dir = folder_dir+folder_name
+            all_matches, all_uri = get_matches(all_matches, all_uri, file_dir)
+
+    all_matches = pd.concat(all_matches, ignore_index=False)
```
nsorros: the default seems to be false for `ignore_index` so why are you explicitly adding this?

merge_results.py
```diff
@@ -0,0 +1,70 @@
...
+    all_uri = pd.concat(all_uri, ignore_index=False)
+    
+    all_matches_url = all_matches.join(
+        all_uri.set_index('Document id'),
```
nsorros: why do you set_index? i don't think this is neccesary

merge_results.py
```diff
@@ -0,0 +1,70 @@
...
+        on='Document id'
+        ) # Join with url
+    
+    all_matches_refs = all_matches_url.join(
+        match_refs.set_index('uber_id'),
```
nsorros: same

merge_results.py
```diff
@@ -0,0 +1,70 @@
...
+    # The name you used in your output file name
+    job_name = "charlene" # "pophealth", "charlene"
+    
+    # The original references data searched for
+    match_refs = pd.read_csv(
+        "/Users/gallaghe/Code/reference-parser/match-references/charlene_publications_format.csv")
+    # "/Users/gallaghe/Code/reference-parser/match-references/MRC_Publications_Nov2018_JGHT_JHSRI_all.csv"
+    # "/Users/gallaghe/Code/reference-parser/match-references/charlene_publications_format.csv"
+
+    folder_dir = './tmp/parser-output/'
```
nsorros: same, this should be an argument because it depends what is the folder of a user

merge_results.py
hblanks: +1 ArgumentParser is your friend. :-)

lizgzil:
@nsorros I've made the changes in the 2 latest commits. There are 2 things I can think of that you might not like:
- having seperate functions for `merge_match_csvs` and `merge_predicted_csvs` even though there are parts of them that do the same thing. I started to write these in one function, but then I thought it might be a bit confusing why we are only taking parts of the data for the case of the predicted_csv_name data.
-
 ```
try:
    pred_data = pd.read_csv(predicted_csv_name)
except:
    print('Read csv issue for file {}'.format(predicted_csv_name))
```
this is because for 4 of the predicted data files an error occured when I tried to read them: 
screenshot 2019-02-13 at 16 19 00

I seem to always have `read_csv` problems 😒. I can show you the problematic files if you want. They are
`dc1b4f8cb79f432a58490dc75866bee6_predicted_reference_structures.csv`
`6a45e70af3548671d48e2e5c57387176_predicted_reference_structures.csv`
`159d0387d70c84eed881c4635549569e_predicted_reference_structures.csv`
`be29f26af5ddde5ffdce2223447449d9_predicted_reference_structures.csv`

File: merge_results.py [hunk 8]
```diff
@@ -0,0 +1,104 @@
...
+
+if __name__ == '__main__':
+
+    file_dir = './tmp/parser-output/charlene' # "pophealth", "charlene"
+    output_name = 'merged_all_matches'
+    
+    match_refs = pd.read_csv(
+        "./match-references/charlene_publications_format.csv")
+    # "./match-references/MRC_Publications_Nov2018_JGHT_JHSRI_all.csv"
+    # "./match-references/charlene_publications_format.csv"
```
nsorros: duplicate

merge_results.py
```diff
@@ -0,0 +1,104 @@
...
+
+    match_csv_names = get_csv_names(file_dir, '_all_match_data')
+    predicted_csv_names = get_csv_names(file_dir, '_predicted_reference_structures')
+
+    all_match = merge_match_csvs(match_csv_names)
```
nsorros: last minute suggestion for the name, concat is probably clearer than merge

merge_results.py
lizgzil: without it the join happens on the original index rather than the document id. Removing it gives this error:
screenshot 2019-02-14 at 11 50 25

merge_results.py
lizgzil: 👍

merge_results.py
lizgzil: 👍

lizgzil:
@nsorros I addressed some of these things and it all runs as expected. Do you want to take a quick look at the latest commit or should I just merge?

PR #58, opened 2019-02-12
Fix an error happening sometime in pdf parsing

SamDepardieu:
# Description

The pdf parser module could in some occasion send different results in
the tests. This commit tries to fix it by setting a specific zoom level.

This PR also changes the `WHO_YEARS` settings, as I came accross it and it was outdated.

## Type of change

# How Has This Been Tested?

Local and unittest (Mac OSx/Python3.6)

### DIFF
--- a/wsf-scraper-web/pdf_parser/pdf_parse.py
+++ b/wsf-scraper-web/pdf_parser/pdf_parse.py
@@ -23,6 +23,8 @@ def parse_pdf_document(document):
         'pdftohtml',
         '-i',
         '-xml',
+        '-zoom',
+        '1.5',
         document.name,
         parsed_path
     ]
--- a/wsf-scraper-web/wsf_scraping/settings.py
+++ b/wsf-scraper-web/wsf_scraping/settings.py
@@ -62,7 +62,7 @@ COOKIES_ENABLED = False
 #  who_iris and who_iris_single_page dedicated settings
 WHO_IRIS_RPP = 250
 WHO_IRIS_LIMIT = False
-WHO_IRIS_YEARS = range(2012, datetime.now().year)
+WHO_IRIS_YEARS = range(2012, datetime.now().year + 1)
 
 # nice dedicated settings
 NICE_GET_HISTORY = False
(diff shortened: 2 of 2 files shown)


wsf-scraper-web/wsf_scraping/settings.py
```diff
@@ -61,7 +62,7 @@
 #  who_iris and who_iris_single_page dedicated settings
 WHO_IRIS_RPP = 250
 WHO_IRIS_LIMIT = False
-WHO_IRIS_YEARS = [2012, 2013, 2014, 2015, 2016, 2017]
+WHO_IRIS_YEARS = range(2012, datetime.now().year)
```
nsorros: this will not include this year, in case you wanted to include which i think we should, otherwise it will scrape and the number of files wont change

hblanks: I tend to agree. @SamDepardieu , should thes be `datetime.now().year + 1) ?`

wsf-scraper-web/pdf_parser/pdf_parse.py
```diff
@@ -23,6 +23,8 @@ def parse_pdf_document(document):
         'pdftohtml',
         '-i',
         '-xml',
+        '-zoom',
+        '1.5',
```
hblanks: Yes! Detailed commit message! Thank you.

wsf-scraper-web/wsf_scraping/settings.py
SamDepardieu: If I remember well, who only publish things from the past year (that's the reason why it was previously set to 2017, not 2018), so I though it wouldn't be useful to set it to `datetime.now().year + 1`. If you think I should, I'm happy to change it, though

nsorros: Better be safe than sorry is my opinion. It is safe to scrape all years so if they change their policy we are covered

SamDepardieu: Ok, agreed

PR #59, opened 2019-02-14
Update regex dictionary

peter337:
Updating the regex dictionary, based on work in https://github.com/wellcometrust/datalabs/pull/188

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -13,8 +13,6 @@ class BaseSettings:
     BLOCKSIZE = 1000
     FUZZYMATCH_THRESHOLD = 0.8
 
-    ORGANISATION = os.environ.get('ORGANISATION', 'nice')
-
     BUCKET = "datalabs-data"
 
     SCRAPER_RESULTS_BASEDIR = "s3://{}/scraper-results".format(BUCKET)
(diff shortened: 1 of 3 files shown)


nsorros:
Good point by @lizgzil. Even better, just remove the organisation from the settings all together and update the documentat that it is not needed. just initialise the `ORGANISATION_REGEX` to the universal regex

settings.py
```diff
@@ -13,8 +13,6 @@ class BaseSettings:
     BLOCKSIZE = 1000
     FUZZYMATCH_THRESHOLD = 0.8
 
-    ORGANISATION = os.environ.get('ORGANISATION', 'nice')
```
nsorros: This will fail if you remove the organisation  `logger.info("[+] Reading input files for %s", settings.ORGANISATION)` but you can remove it so it is fine. Make sure the code runs and update the documentation to say that the user does not need to initialise ORGANISATION

settings.py
```diff
@@ -13,10 +13,12 @@ class BaseSettings:
     BLOCKSIZE = 1000
     FUZZYMATCH_THRESHOLD = 0.8
 
+    ORGANISATION = os.environ.get('ORGANISATION', 'nice')
```
nsorros: why is this back in?

peter337: Whoops! I'll remove that

PR #65, opened 2019-02-14
Fix #43 for when no reference components are found

### LINKED ISSUE #43: Fix predict for when 'Predicted Category' is not found
I came across an error in `predict` where `predict_all['Predicted Category']` was not found, see screenshot:

screenshot 2019-01-31 at 11 44 49

Perhaps worth checking whether `reference_components` is ever empty, since it shouldn't be, but this could give rise to no 'Predicted Category' field.

lizgzil:
# Description

On very rare occassions there are no reference components found when the references section is cleaned and split, and therefore the code fails because in `predict.py` there is no `predict_all['Predicted Category']` found.

This fixes the issue, but in the future we would like to rather than just skip processing the document, return empty lists in predict_references, predict_structure and fuzzy_match_blocks.

## Type of change

### DIFF
--- a/refparse.py
+++ b/refparse.py
@@ -121,6 +121,11 @@ def run_predict(scraper_file, references_file,
         # Split references into components
         splitted_components = process_references(splitted_references)
 
+        # TO DO: Rather than just skip,
+        # return empty lists in predict_references, predict_structure and fuzzy_match_blocks
+        if len(splitted_components)==0:
+            continue
+
         # Predict the references types (eg title/author...)
         # logger.info('[+] Predicting the reference components')
         components_predictions = predict_references(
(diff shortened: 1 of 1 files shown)


nsorros:
I will defer to @hblanks on this.

refparse.py
```diff
@@ -121,6 +121,11 @@ def run_predict(scraper_file, references_file,
         # Split references into components
         splitted_components = process_references(splitted_references)
 
+        # TO DO: Rather than just skip,
+        # return empty lists in predict_references, predict_structure and fuzzy_match_blocks
+        if len(splitted_components)==0:
```
hblanks: Thanks for the fix! I think you want spaces between your boolean operators. I.e. `x == 0`.

It's also actually OK to just do `if not splitted_components:`.

lizgzil: 👍  I've done this in the latest commit

lizgzil: actually I realised it indeed needs to be `if len(splitted_components) == 0:`

PR #76, opened 2019-02-19
Fix the codepipeline related commands in the makefile

SamDepardieu:
# Description

The new dev environment changed the path and made it impossible for codebuild to run the make docker-test recipe.
This pr fixes this by changing the path in the buidspec file and by changing the test running command.
Also update the dependencies.

## Type of change


policytool/Makefile
```diff
@@ -54,7 +54,7 @@ virtualenv: $(VIRTUALENV)/.installed
 docker-test: docker-build
 	docker run -v $$(pwd)/requirements.txt:/requirements.txt \
 	    --rm $(ECR_IMAGE):$(VERSION) \
-		sh -c "pip3 install -r requirements.txt && python3 -m unittest"
+		sh -c "cd /src/policytool/scraper && python3 -m unittest"
```
lizgzil: is it `cd /src/policytool/scraper` and not `cd ./src/policytool/scraper` then? (I always do the latter but not sure if the former works the same)

SamDepardieu: In this case, we're using `cd` inside the Docker container. The path to access our folder is from the root, so the correct path is `/src/policytool/scraper`. If we wanted to cd inside the same folder in our current path, then we would use `./src/policytool/scraper`. :)

lizgzil: 👍 thanks

PR #77, opened 2019-02-19
Commit the setup.py file that was previously ignored

SamDepardieu:
We need it if we want it to build on AWS

### DIFF
--- a/policytool/setup.py
+++ b/policytool/setup.py
@@ -8,7 +8,7 @@ setuptools.setup(
     name="wellcome-policytool",
     version="0.0.1",
     author="Wellcome Trust Data Labs Team",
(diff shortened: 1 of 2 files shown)


policytool/setup.py
```diff
@@ -0,0 +1,43 @@
...
+
+# Current directory must be data-labs repo before running setup.py!
+with open("README.md", "r") as f:
+    long_description = f.read()
+
+setuptools.setup(
+    name="wellcome-policytool",
+    version="0.0.1",
+    author="Wellcome Trust Data Labs Team",
+    author_email=" datalabs-engineering@wellcomecloud.onmicrosoft.com",
```
lizgzil: is there supposed to be a space before the email?

SamDepardieu: Nope

PR #83, opened 2019-02-22
Fix the Nice spider after their site re-design

SamDepardieu:
# Description

* Remove the ajax request as it doesn't exist anymore
* Add a "next page" button yield as it now have pages instead of an
auto-loader
* Change start url to match all published guidances
* We're not scraping text anymore, but were looking if the item had
text to scrape it
 * Changes the text attribute to has_text
 * We still know if the pdf_parser was able to proces the pdf without
relying on text
 * Fix the tests to work with the new version of the settings file
 * Make database url environ a `.get`, so importing the settings won't crash. The laoding will still crash after if no database is state, so it won't be a quiet crash anyway.

## Type of change

- [x] :bug: Bug fix


policytool/scraper/wsf_scraping/spiders/nice_spider.py
```diff
@@ -126,11 +120,12 @@ def parse_article(self, response):
...
 
+        date = response.css(
+            '.published-date time::attr(datetime)'
+        ).extract_first()
         data_dict = {
             'title': response.css('h1::text').extract_first(),
-            'year': response.css(
-                '.published-date time::attr(datetime)'
-            ).extract_first()[:4]
+            'year': date[:4] if date else None
```
hblanks: wouldn't be bad to pull the full date eventually, but this looks like a good start in any case

PR #85, opened 2019-02-25
Increase clarity in sample code in README

peter337:
Simple changes to the README to make it clearer where changes are required by the reader.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ mkdir -p ./tmp/parser-output/output_folder_name
 
 python ./refparse.py \
     --scraper-file "s3://datalabs-data/scraper-results/msf/20190117.json" \
-    --references-file "match-references/MRC_Publications_Nov2018_JGHT_JHSRI.csv" \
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,7 +66,7 @@ mkdir -p ./tmp/parser-output/output_folder_name
 
 python ./refparse.py \
     --scraper-file "s3://datalabs-data/scraper-results/msf/20190117.json" \
-    --references-file "match-references/MRC_Publications_Nov2018_JGHT_JHSRI.csv" \
+    --references-file "path/to/references.csv" \
```
lizgzil: I will defer to @hblanks or @SamDepardieu to say whether this is the correct way to write a generic file path. I like it though

peter337: Something I've seen on Stack Exchange a bunch, so nicked it from there.

hblanks: yes, this is a common idiom, so looks like a good choice

PR #87, opened 2019-02-28
Transition from model and vectorizer to pipeline

nsorros:
This PR moves away from having two separate pickle files for a model and a vectorizer and utilises the pipeline [functionality](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html) of sklearn to combine the two so that we do not need to carry around two objects as well as pass two arguments. Instead all the logic that comes from training the model is incorporates into the pipeline object that accepts a list of strings and returns a classification and a probablity.


peter337:
Should this also include changes to parse_latest.py? e.g. line 68, ```args.vectorizer_file```.

nsorros:
@peter337 good point, will make the necessary adjustments

README.md
```diff
@@ -87,9 +86,10 @@ If you want to specify the arguments for the other inputs then you can, otherwis
 ```
 python ./parse_latest.py msf \
     --references-file "path/to/references.csv" \
-    --model-file "s3://datalabs-data/reference_parser_models/RefSorter_classifier.pkl" \
-    --vectorizer-file "s3://datalabs-data/reference_parser_models/RefSorter_vectorizer.pkl" \
+    --model-file "s3://datalabs-data/reference_parser_models/
+    reference_parser_pipeline.pkl" \
+
```
peter337: Think that lines 89 and 90 should be a single line, and that line 92 should be deleted.

nsorros: 👍

PR #90, opened 2019-03-01
Fix examples by adding a correct path to publications so they run

nsorros:
This PR aims to make the examples in the readme run. To do so I add an s3 url that is valid for Wellcome publications and fix the path of merge results which does not use the `file://` prefix.

This is a temporary solution until we add some sample data into the repo so that a user that does not work at wellcome can run the code

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ mkdir -p ./tmp/parser-output/output_folder_name
 
 python ./refparse.py \
     --scraper-file "s3://datalabs-data/scraper-results/msf/20190117.json" \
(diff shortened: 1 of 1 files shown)


peter337:
Out of interest, has/will Aoife see(n) this file? Would be interested to hear her perspective on how this README reads.

nsorros:
@peter337 good point but i am not sure how much she can add given she is not a programmer. i can add her to the PR but cannot seem to find her handle.

nsorros:
@aoifespenge if you have time, have a look as per peter suggestion

PR #91, opened 2019-03-01
Scraper download only

SamDepardieu:
# Description

As part of the airflow transformation, this PR introduce a breaking change to the scraper.
 - The scraper now only download the PDF and write to a manifest file in the given destination
 - The pdf_parser is now a separate module (broken at the moment)
 - The default behaviour of the scraper is now to download all pdf
 - The scraper comes with two different storage way: locally and on s3

## Type of change

- [x] :fire: Breaking change

# How Has This Been Tested?

Mac OSx 10.14.2 / Python 3.6:
 - Local runs
 - Unit testing


policytool/pdf_parser/main.py
```diff
@@ -0,0 +1,148 @@
+from pdf_parser.pdf_parse import parse_pdf_document, grab_section
+from scraper.utils.dbTools import DatabaseConnector
+import os
+import datetime
+import logging
```
hblanks:
```
import os
import datetime
import logging

from pdf_parser.pdf_parse import ...
from scraper.utils.dbTools import ...
```

Per https://www.python.org/dev/peps/pep-0008/#imports

policytool/pdf_parser/main.py
```diff
@@ -0,0 +1,148 @@
...
+
+logger = logging.getLogger(__name__)
+logger.setLevel('INFO')
+
+
+KEYWORDS_FILE = ''
+SECTIONS_KEYWORDS_FILE = ''
+KEYWORDS_CONTEXT = ''
```
hblanks: Sorry, what are these empty string constants about? I feel like calling `parse_keywords_file()` with an empty string and using the try block to get back `[]` is a bit odd. Maybe there's a better way?

Related: as all capitalized variables, I'm assuming the standard Python convention that these variables are constants and won't be reassigned elsewhere. But, that also makes me not sure why have them or have them be empty strings. Maybe `None`?

policytool/pdf_parser/main.py
```diff
@@ -0,0 +1,148 @@
...
+            logger.debug("Successfully opened %s", file_path)
+            for line in f:
+                line = line.replace('\n', '')
+                if line and line[0] != '#':
+                    keywords_list.append(line.lower())
+    except IOError:
+        logger.warning(
+            "Unable to open keywords file at location %s", file_path
+        )
+        keywords_list = []
```
hblanks: In production, I suspect we'd want to get exception reports whenever we're not able to read a file. Perhaps this try/except block should be removed and the code refactored.

policytool/pdf_parser/main.py
```diff
@@ -0,0 +1,148 @@
...
+    # Remove the path from the value we are storing
+    item['pdf'] = os.path.basename(item['pdf'])
+    item['provider'] = spider_name
+    item['date_scraped'] = datetime.now().strftime('%Y%m%d')
+    item['has_text'] = True
+    return item
+
+
+if __name__ == '__main__':
+    item = {}
```
hblanks: wouldn't be bad to set up an ArgumentParser here, so that `-h` prints help based on `__doc__`. I.e.
```
parser = argparse.ArgumentParser(__doc__.strip())
parser.parse_args()
```

policytool/pdf_parser/main.py
```diff
@@ -0,0 +1,148 @@
...
+    name = ''
+    scraped = False
+    database = DatabaseConnector()
+    if not scraped:
```
hblanks: `not scraped` will always be true.

policytool/scraper/feed_storages/ManifestFeedStorage.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class ManifestFeedStorage(BlockingFeedStorage):
+    """Stores Scrapy feed results as json files in S3.
+    """
+
+    def __init__(self, uri):
+        """Initialise the Feed Storage, giving it AWS informations."""
+        self.logger = logging.getLogger(__name__)
+        self.parsed_url = urlparse(uri)
```
hblanks: Good call checking your inputs on init!

policytool/scraper/feed_storages/ManifestFeedStorage.py
```diff
@@ -0,0 +1,28 @@
...
+from six.moves.urllib.parse import urlparse
+from scrapy.extensions.feedexport import BlockingFeedStorage
+
+
+class ManifestFeedStorage(BlockingFeedStorage):
+    """Stores Scrapy feed results as json files in S3.
+    """
+
+    def __init__(self, uri):
+        """Initialise the Feed Storage, giving it AWS informations."""
```
hblanks: This storage doesn't seem to be given AWS informations. Maybe you mean:

```
Creates a storage instance for writing PDF manifests to either local file or S3 urls.
```

policytool/scraper/feed_storages/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .AWSFeedStorage import AWSFeedStorage
+from .ManifestFeedStorage import ManifestFeedStorage
+
+__all__ = [AWSFeedStorage, ManifestFeedStorage]
```
hblanks: I'd probably forego this file and just let the calling code import what it needs. (It also seems like the caller only needs to import ManifestFeedStorage)

policytool/scraper/wsf_scraping/storage.py
```diff
@@ -0,0 +1,158 @@
...
+                current_manifest[item['hash'][:2]] = [item['hash']]
+
+        self.client.put_object(
+            Bucket=self.bucket,
+            Prefix=self.prefix,
+            Key='policytool-scrape--scraper-{organisation}.json'.format(
+                organisation=self.organisation,
+            ),
+            Body=current_manifest.encode('utf-8')
+        )
```
hblanks: It appears we're downloading from S3 and writing to S3 every time we get a new PDF.

Perhaps we should be caching within the module, so that we only write for every new PDF?

Or, is there a way to write the manifest in one go at the end, since it probably fits in memory? I get the impression that we'll call `close_spider()` on the pipeline at the end...that could be a good place for writing the manifest. (And in that case, you could just have `get_manifest()` and `put_manifest()`.

policytool/pdf_parser/main.py
SamDepardieu: This is the main file for the pdf parser module. I'm fixing it in another branch, because I had to re-write it anyway for it to be used as a separate task. So at the moment this is just some unused code waiting for a re-write.

policytool/pdf_parser/main.py
SamDepardieu: Same as above, it's a pdf parser issue. I'm taking good note though, will address in the next branch 👍

policytool/pdf_parser/main.py
SamDepardieu: 👍

policytool/pdf_parser/main.py
SamDepardieu: Yup, same as ubove, dummy code

policytool/pdf_parser/main.py
SamDepardieu: I should have commented it, though.

policytool/scraper/feed_storages/ManifestFeedStorage.py
SamDepardieu: Updated. Also added a bit of docs about how the feed storage actually works. 👍

policytool/scraper/wsf_scraping/file_system.py
```diff
@@ -0,0 +1,158 @@
...
+            - data_file: The file object sent by Scrapy's feed storage.
+        """
+        return
+
+
+class S3FileSystem(FileSystem):
+    def __init__(self, path, organisation):
+        self.client = boto3.client('s3')
+        self.bucket = path.split('/')[0]
+        self.prefix = path.split('/')[1:]
```
hblanks:
```
self.bucket, self.prefix = path.split('/', 1)
```

PR #97, opened 2019-03-05
Refactor predict structure

nsorros:
This is one of the biggest refactor to date so please do review and call out any problems.

This refactor aims to achieve a couple of goals
* Remove use of pandas from predict (now named parse)
* Write tests for predict (now named parse)
* Rename variables, functions and files to make the code more readable and names more consistent
* Remove the need to document id and reference id information inside predict (now parse)
* Introduce a high level parse function called structure_reference that a accept a reference and returnes structured references
* Refactor functions in predict so that they work in one reference at at time and the code in refparse to deal with that
* Silence logger to make what is printed out less noisy but still informative

This is a big change but one that I wanted us to do for a couple of months. As a side effect of that change the code runs ~20x faster. Weirdly when I reintroduced pool map in my last commit it slowed things down so I would like to understand why with you @hblanks or @SamDepardieu.

Finally, the idea is that all these changes do move us towards the right direction so if you have second thoughts about any of the above, please do raise them.

### DIFF
--- a/refparse.py
+++ b/refparse.py
@@ -13,10 +13,8 @@ import sentry_sdk
 
 from utils import (FileManager,
                    FuzzyMatcher,
-                   process_references_section,
-                   process_references,
-                   predict_references,
-                   predict_structure)
+                   split_references_section,
+                   structure_references)
 from models import DatabaseEngine
 from settings import settings
 
@@ -118,61 +116,32 @@ def run_predict(scraper_file, references_file,
             nb_documents
         ))
 
-        # Split references section into references
-        splitted_references = process_references_section(
+        splitted_references = split_references_section(
             doc,
             settings.ORGANISATION_REGEX
         )
 
-        # Split references into components
-        splitted_components = process_references(splitted_references)
-
-        # TO DO: Rather than just skip,
-        # return empty lists in predict_references, predict_structure and fuzzy_match_blocks
-        if len(splitted_components) == 0:
-            continue
-
-        # Predict the references types (eg title/author...)
-        # logger.info('[+] Predicting the reference components')
-        components_predictions = predict_references(
+        structured_references = structure_references(
             pool_map,
             model,
-            splitted_components['Reference component']
+            splitted_references
         )
 
-        # Link predictions back with all original data (Document id etc)
-        # When we merge and splitted_components is a dict not a dataframe then
-        # we could just merge the list of dicts
-        reference_components_predictions = splitted_components
-        reference_components_predictions["Predicted Category"] = [
-            d["Predicted Category"] for d in components_predictions
-        ]
-        reference_components_predictions["Prediction Probability"] = [
-            d["Prediction Probability"] for d in components_predictions
-        ]
-
-        # Predict the reference structure
-        predicted_reference_structures = predict_structure(
-            pool_map,
-            reference_components_predictions,
-            settings.PREDICTION_PROBABILITY_THRESHOLD
-        )
-
-        all_match_data = fuzzy_matcher.fuzzy_match(
-            predicted_reference_structures
+        matched_references = fuzzy_matcher.fuzzy_match(
+            structured_references
         )
 
         if output_url.startswith('file://'):
             # use everything after first two slashes; this handles
             # absolute and relative urls equally well
             output_dir = output_url[7:]
-            predicted_reference_structures.to_csv(
+            structured_references.to_csv(
                 os.path.join(
                     output_dir,
                     f"{doc.id}_{settings.PREF_REFS_FILENAME}"
                     )
                 )
-            all_match_data.to_csv(
+            matched_references.to_csv(
                 os.path.join(
                     output_dir,
                     f"{doc.id}_{settings.MATCHES_FILENAME}"
@@ -181,8 +150,8 @@ def run_predict(scraper_file, references_file,
         else:
             database = DatabaseEngine(output_url)
             database.save_to_database(
-                predicted_reference_structures,
-                all_match_data,
+                structured_references,
+                matched_references,
             )
         nb_references += len(splitted_references)
 
--- a/refparse.py
+++ b/refparse.py
@@ -15,7 +15,7 @@ import pandas as pd
 from utils import (FileManager,
                    FuzzyMatcher,
                    split_section,
-                   structure_references)
+                   structure_reference)
 from models import DatabaseEngine
 from settings import settings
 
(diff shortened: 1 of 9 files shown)


refparse.py
```diff
@@ -109,70 +126,51 @@ def run_predict(scraper_file, references_file,
...
-        components_predictions = predict_references(
-            pool_map,
-            model,
-            splitted_components['Reference component']
+        # For some weird reason not using pool map 
+        #   in my laptop is more performant
+        structured_references = pool_map(
+            # A better name would be parse_reference
+            #   here but we use it differently here
+            partial(structure_reference, model),
```
hblanks: nice use of partial!

utils/__init__.py
```diff
@@ -1,17 +1,17 @@
...
+    merge_components,
+    structure_reference,
     FuzzyMatcher,
     FileManager,
     serialise_matched_reference,
-    serialise_reference,
-    predict_references,
-    predict_structure,
-    split_section,
+    serialise_reference
```
hblanks: No need to change this, but as an aside, if your module structure is leading you to have to update `__all__` with some frequency, it usually suggests you want a flatter module structure. E.g. here:

```
utils.py
refererence/split.py
reference/predict.py
reference/parse.py
reference/serialize.py  # US orthography? Up to you but Python stdlib does use z's, if it maters.
fuzzymatch.py
filemanager.py
```

The biggest benefit isn't that you don't have to update `__init__.py` but rather that when you read, you can easily tell where a function is defined, just by the module names. For similar reasons we sometimes prefer plain `import` over `from X import Y`.

Anyway, just food for thought, nothing you should change in this PR.

utils/parse.py
```diff
@@ -0,0 +1,141 @@
...
+        # Notice that we remove spaces for elem
+        elem.strip()
+        for elem in reference.replace(
+            ',', '.'
+        ).replace(
+            '?', '?.'
+        ).replace(
+            '!', '!.'
+        ).split(".")
+    ]
```
hblanks: A regex `re.compile('[,?!]')` replace would go through your strings once instead of 3 times. Not a big deal but FYI.

utils/parse.py
```diff
@@ -0,0 +1,141 @@
...
+        ]
+        if category_components:
+            max_prob_component = max(category_components, key=lambda x: x['Prediction Probability'])
+            max_prob_group = max_prob_component['Group']
+
+            group_components = [
+                comp for comp in category_components
+                if comp['Group'] == max_prob_group
+            ]
+            merged_component = ", ".join([
```
hblanks: not a big deal, but the `[` in this line is unnecessary. Anything inside `()` can act as  a generator expression. E.g.:

```
'|'.join(str(i) for i in range(5))
```

utils/parse.py
```diff
@@ -0,0 +1,141 @@
...
+                comp['Reference component'] for comp in group_components
+            ])
+            
+        structured_reference.update({category: merged_component})
+
+    return structured_reference
+
+
+def is_year(component):
+    valid_years_range = range(1800, 2020)
```
hblanks: You may want to replace 2020 with a constant. And use `<` and `>`:

```python
# outside function
VALID_YEARS = (1800, datetime.date.today().year + 1)

# inside function
return component.isdecimal() and VALID_YEARS[0] <= int(component) < VALID_YEARS[1]
```

utils/fuzzymatch.py
```diff
@@ -64,10 +63,10 @@ def match_vectorised(self, predicted_publications):
...
 
     def fuzzy_match(self, predicted_publications):
-        self.logger.info(
-            "Fuzzy matching for %s predicted publications ...",
-            len(predicted_publications)
-        )
+        # self.logger.info(
+        #     "Fuzzy matching for %s predicted publications ...",
+        #     len(predicted_publications)
+        # )
```
SamDepardieu: If you're not using it, just remove it ;)

utils/parse.py
```diff
@@ -0,0 +1,141 @@
...
+    - model: The trained multinomial naive Bayes model for predicting the categories of reference components
+    - reference_components: A list of reference components
+    Output:
+    - A list of dicts [{"Predicted Category": , "Prediction Probability": } ...]
+    """
+
+    # logger.info(
+    #     "[+] Predicting the categories of %s  reference components ...",
+    #     str(len(reference_components))
+    # )
```
SamDepardieu: Same, if it's commented, you can delete it

utils/parse.py
```diff
@@ -0,0 +1,141 @@
...
+            cat = component_predictions[i]
+            prob = component_predictions_probs[i]
+
+        predicted_components.append({
+            'Reference component': component,
+            'Predicted Category': cat,
+            'Prediction Probability': prob
+        })
+
+    # logger.info("Predictions complete")
```
SamDepardieu: Same again

peter337:
Runs fine on my machine - but unfortunately hasn't magically solved the problem of references being dropped!

refparse.py
```diff
@@ -109,70 +126,51 @@ def run_predict(scraper_file, references_file,
...
+    for i, doc in enumerate(sectioned_documents):
+        logger.info('[+] Processing references from document {}'.format(
+            i
         ))
 
-        # Split references section into references
-        splitted_references = process_references_section(
-            doc,
+        splitted_references = split_section(
+            doc.section,
```
peter337: In the interest of using minimally sufficient pandas, I think this should be doc['section'] instead?
Same comment on lines 155,156 below.

nsorros: doc.section is used a name tuple so no use of pandas here :)

refparse.py
```diff
@@ -109,70 +126,51 @@ def run_predict(scraper_file, references_file,
         settings.FUZZYMATCH_THRESHOLD
     )
 
+    sectioned_documents = transform_scraper_file(scraper_file)
+
     t0 = time.time()
     nb_references = 0
-    nb_documents = sum(scraper_file['sections'].notnull())
-    for i, doc in enumerate(transform_scraper_file(scraper_file)):
-        logger.info('[+] Processing references from document {} of {}'.format(
```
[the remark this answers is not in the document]
nsorros: The problem is that the solution you provide is not performant, i can explain if you want

hblanks:
@nsorros  - I think this broke the build. Logs up at https://eu-west-1.console.aws.amazon.com/codesuite/codebuild/projects/datalabs-airflow/build/policytool%3A5894c697-4de0-4ac8-8871-65207e5783f8/log?region=eu-west-1# indicate new tests are failing.

PR #100, opened 2019-03-10
Add hard text search to Policy Tool

peter337:
Won't be available Monday morning so adding this now. Currently runs, but has a few bugs.

@nsorros, I've encountered a bug while adding rows to `matched_references` which seems to be caused by its duplicate columns name - three of the columns are called 'index'. I currently work around that by removing columns with duplicate names on line 60 of hard_text_search.py.
Do you know what the role if these multiple index columns is/is there any issue in removing them?

### DIFF
--- a/refparse.py
+++ b/refparse.py
@@ -98,20 +98,20 @@ def get_file(file_str, file_type, get_scraped=False):
 def remove_dups_and_concat(fuzzy_matches, text_matches):
     #Remove duplicate columns and short matches in the fuzzy matches
     fuzzy_matches = fuzzy_matches.loc[:,~fuzzy_matches.columns.duplicated()]
-    fuzzy_matches = fuzzy_matches.loc[fuzzy_matches['WT_Ref_Title'].str.len() >= settings.FUZZY_MATCH_CHAR_LIMIT]
+    fuzzy_matches = fuzzy_matches.loc[fuzzy_matches['WT_Ref_Title'].str.len() >= settings.MIN_CHAR_LIMIT]
 
-    if not text_matches.empty:
-        duplicate_matches = fuzzy_matches['WT_Ref_Id'][fuzzy_matches['WT_Ref_Id'].isin(text_matches['WT_Ref_Id'])]
-
-        #For duplicate matches: remove from text_matches, and renames 'Match_algorithm' in fuzzy_matches
-        text_matches = text_matches[~text_matches['WT_Ref_Id'].isin(duplicate_matches)]
-        fuzzy_matches['Match_algorithm'][fuzzy_matches['WT_Ref_Id'].isin(duplicate_matches)] = "Fuzzy and Text Searches"
-
-        all_matches = pd.concat([fuzzy_matches, text_matches])
-        return (all_matches)
+    if text_matches.empty:
+        return fuzzy_matches
 
     else:
-        return (fuzzy_matches)
+        duplicate_ref_ids = fuzzy_matches['WT_Ref_Id'][fuzzy_matches['WT_Ref_Id'].isin(text_matches['WT_Ref_Id'])]
+
+        #For duplicate matches: remove from text_matches, and renames 'Match_algorithm' in fuzzy_matches
+        text_matches = text_matches[~text_matches['WT_Ref_Id'].isin(duplicate_ref_ids)]
+        fuzzy_matches['Match_algorithm'][fuzzy_matches['WT_Ref_Id'].isin(duplicate_ref_ids)] = "Fuzzy and Text Searches"
+
+        all_matches = pd.concat([fuzzy_matches, text_matches])
+        return all_matches
 
 
 def run_predict(scraper_file, references_file,
--- a/refparse.py
+++ b/refparse.py
@@ -102,14 +102,16 @@ def remove_dups_and_concat(fuzzy_matches, text_matches):
 
     if not text_matches.empty:
         duplicate_matches = fuzzy_matches['WT_Ref_Id'][fuzzy_matches['WT_Ref_Id'].isin(text_matches['WT_Ref_Id'])]
-
+        print(duplicate_matches)
         #For duplicate matches: remove from text_matches, and renames 'Match_algorithm' in fuzzy_matches
         text_matches = text_matches[~text_matches['WT_Ref_Id'].isin(duplicate_matches)]
-        fuzzy_matches['Match_algorithm'][fuzzy_matches['WT_Ref_Id'].isin(duplicate_matches)] = "Fuzzy Matcher and Hard Text"
+        fuzzy_matches['Match_algorithm'][fuzzy_matches['WT_Ref_Id'].isin(duplicate_matches)] = "Fuzzy and Text Searches"
 
-    all_matches = pd.concat([fuzzy_matches, text_matches])
+        all_matches = pd.concat([fuzzy_matches, text_matches])
+        return (all_matches)
 
-    return(all_matches)
+    else:
+        return (fuzzy_matches)
 
 
 def run_predict(scraper_file, references_file,
--- a/refparse.py
+++ b/refparse.py
@@ -109,7 +109,7 @@ def remove_dups_and_concat(fuzzy_matches, text_matches):
 
     #For duplicate matches: remove from text_matches, and renames 'Match_algorithm' in fuzzy_matches
     text_matches = text_matches[~text_matches['WT_Ref_Id'].isin(duplicate_ref_ids)]
-    fuzzy_matches['Match_algorithm'][fuzzy_matches['WT_Ref_Id'].isin(duplicate_ref_ids)] = "Fuzzy and Text Searches"
+    fuzzy_matches.at[fuzzy_matches['WT_Ref_Id'].isin(duplicate_ref_ids), 'Match_algorithm'] = "Fuzzy and Text Searches"
 
     all_matches = pd.concat([fuzzy_matches, text_matches])
     return all_matches
--- a/refparse.py
+++ b/refparse.py
@@ -17,7 +17,7 @@ from utils import (FileManager,
                    FuzzyMatcher,
                    split_section,
                    structure_reference,
-                   clean_series_text,
+                   clean_text,
                    hard_text_search)
 from models import DatabaseEngine
 from settings import settings
@@ -120,10 +120,15 @@ def run_predict(scraper_file, references_file,
     ref_file = get_file(references_file, 'csv')
     check_references_file(ref_file, references_file)
 
-    clean_ref_file = pd.DataFrame({
-        'uber_id' : ref_file['uber_id'],
-        'title' : clean_series_text(ref_file['title'])
-    })
+    #Minimum no. of characters to allow in a title, informed by shortest true positive in UKRI data
+    min_chars = 18
+    #Duplicates the references file and cleans it up for the hard text search
+    clean_ref_file = ref_file
+    clean_ref_file['title'] = [clean_text(x) for x in clean_ref_file['title']]
+    clean_ref_file = clean_ref_file.loc[clean_ref_file['title'].str.len() >= min_chars]
+    clean_ref_file = clean_ref_file.drop_duplicates(subset = 'title')
+    #Converts to dict, for significantly faster run time over pandas DF
+    clean_ref_file = clean_ref_file.to_dict(orient = 'list')
     
     # Loading the pipeline
     model = get_file(model_file, 'pickle')
@@ -163,14 +168,14 @@ def run_predict(scraper_file, references_file,
             doc.uri
         )
 
-        matched_references = fuzzy_matcher.fuzzy_match(
+        fuzzy_matched_references = fuzzy_matcher.fuzzy_match(
             structured_references
         )
 
-        matched_references = hard_text_search(
+        all_matched_references = hard_text_search(
             doc,
             clean_ref_file,
-            matched_references
+            fuzzy_matched_references
         )
 
         if output_url.startswith('file://'):
@@ -183,7 +188,7 @@ def run_predict(scraper_file, references_file,
                     f"{doc.id}_{settings.PREF_REFS_FILENAME}"
                     )
                 )
-            matched_references.to_csv(
+            all_matched_references.to_csv(
                 os.path.join(
                     output_dir,
                     f"{doc.id}_{settings.MATCHES_FILENAME}"
@@ -193,7 +198,7 @@ def run_predict(scraper_file, references_file,
             database = DatabaseEngine(output_url)
             database.save_to_database(
                 structured_references,
-                matched_references,
+                all_matched_references,
             )
         nb_references += len(splitted_references)
 
--- a/refparse.py
+++ b/refparse.py
@@ -103,15 +103,14 @@ def remove_dups_and_concat(fuzzy_matches, text_matches):
     if text_matches.empty:
         return fuzzy_matches
 
-    else:
-        duplicate_ref_ids = fuzzy_matches['WT_Ref_Id'][fuzzy_matches['WT_Ref_Id'].isin(text_matches['WT_Ref_Id'])]
+    duplicate_ref_ids = fuzzy_matches['WT_Ref_Id'][fuzzy_matches['WT_Ref_Id'].isin(text_matches['WT_Ref_Id'])]
 
-        #For duplicate matches: remove from text_matches, and renames 'Match_algorithm' in fuzzy_matches
-        text_matches = text_matches[~text_matches['WT_Ref_Id'].isin(duplicate_ref_ids)]
-        fuzzy_matches['Match_algorithm'][fuzzy_matches['WT_Ref_Id'].isin(duplicate_ref_ids)] = "Fuzzy and Text Searches"
+    #For duplicate matches: remove from text_matches, and renames 'Match_algorithm' in fuzzy_matches
+    text_matches = text_matches[~text_matches['WT_Ref_Id'].isin(duplicate_ref_ids)]
+    fuzzy_matches['Match_algorithm'][fuzzy_matches['WT_Ref_Id'].isin(duplicate_ref_ids)] = "Fuzzy and Text Searches"
 
-        all_matches = pd.concat([fuzzy_matches, text_matches])
-        return all_matches
+    all_matches = pd.concat([fuzzy_matches, text_matches])
+    return all_matches
 
 
 def run_predict(scraper_file, references_file,
--- a/refparse.py
+++ b/refparse.py
@@ -17,8 +17,7 @@ from utils import (FileManager,
                    FuzzyMatcher,
                    split_section,
                    structure_reference,
-                   clean_text,
-                   hard_text_search)
+                   HardTextSearch)
 from models import DatabaseEngine
 from settings import settings
 
@@ -119,16 +118,6 @@ def run_predict(scraper_file, references_file,
     # Loading the references file
     ref_file = get_file(references_file, 'csv')
     check_references_file(ref_file, references_file)
-
-    #Minimum no. of characters to allow in a title, informed by shortest true positive in UKRI data
-    min_chars = 18
-    #Duplicates the references file and cleans it up for the hard text search
-    clean_ref_file = ref_file
-    clean_ref_file['title'] = [clean_text(x) for x in clean_ref_file['title']]
-    clean_ref_file = clean_ref_file.loc[clean_ref_file['title'].str.len() >= min_chars]
-    clean_ref_file = clean_ref_file.drop_duplicates(subset = 'title')
-    #Converts to dict, for significantly faster run time over pandas DF
-    clean_ref_file = clean_ref_file.to_dict(orient = 'list')
     
     # Loading the pipeline
     model = get_file(model_file, 'pickle')
@@ -136,6 +125,11 @@ def run_predict(scraper_file, references_file,
     fuzzy_matcher = FuzzyMatcher(
         ref_file,
         settings.FUZZYMATCH_THRESHOLD
+    )
+
+    text_matcher = HardTextSearch(
+        ref_file,
+        40
     )
 
     sectioned_documents = transform_scraper_file(scraper_file)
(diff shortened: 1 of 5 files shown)


lizgzil:
> Won't be available Monday morning so adding this now. Currently runs, but has a few bugs.
> 
> @nsorros, I've encountered a bug while adding rows to `matched_references` which seems to be caused by its duplicate columns name - three of the columns are called 'index'. I currently work around that by removing columns with duplicate names on line 60 of hard_text_search.py.
> Do you know what the role if these multiple index columns is/is there any issue in removing them?

I think the index issue is because of the pd.concat with .reset_index() on line 50 of fuzzymatch. I am quite sure it is fine to deduplicate the columns like you say. I have a memory of fixing this, but can't find where. I think when we start to output to JSON it becomes more of a problem.

refparse.py
```diff
@@ -118,6 +120,11 @@ def run_predict(scraper_file, references_file,
     ref_file = get_file(references_file, 'csv')
     check_references_file(ref_file, references_file)
 
+    clean_ref_file = pd.DataFrame({
```
nsorros: i know it is a WIP but some thoughs as you develop this might be helpful. This is not needed here. Best to deal with this problem inside the hard_test_search. Pass the section there and clean it when needed.

refparse.py
```diff
@@ -160,6 +167,12 @@ def run_predict(scraper_file, references_file,
             structured_references
         )
 
+        matched_references = hard_text_search(
```
nsorros: this is the same name as the fuzzy matcher now returns so you might need to work around the names a bit to differentiate. additionally do not forget to deduplicate, somwhere.

peter337:
@nsorros @lizgzil I think this now runs as fast as I can get it, so removing the WIP tag and opening it for review.
For info, I've compared the run time with and without the Hard Text search on the who_iris documents. Without the Hard Text search, it takes 0.8s/document, and with it takes 16.5s/document (still down from the 50s/doc we had with pandas!).

lizgzil:
> @nsorros @lizgzil I think this now runs as fast as I can get it, so removing the WIP tag and opening it for review.
> For info, I've compared the run time with and without the Hard Text search on the who_iris documents. Without the Hard Text search, it takes 0.8s/document, and with it takes 16.5s/document (still down from the 50s/doc we had with pandas!).

Great. Will take a look tomorrow, hopefully we can get the time down, but I guess it's not that surprising it should take so much longer :(

refparse.py
```diff
@@ -163,14 +168,14 @@ def run_predict(scraper_file, references_file,
             doc.uri
         )
 
-        matched_references = fuzzy_matcher.fuzzy_match(
+        fuzzy_matched_references = fuzzy_matcher.fuzzy_match(
```
nsorros: why did we go back to the previous names?

refparse.py
```diff
@@ -118,6 +120,16 @@ def run_predict(scraper_file, references_file,
...
 
+    #Minimum no. of characters to allow in a title, informed by shortest true positive in UKRI data
+    min_chars = 18
+    #Duplicates the references file and cleans it up for the hard text search
+    clean_ref_file = ref_file
+    clean_ref_file['title'] = [clean_text(x) for x in clean_ref_file['title']]
+    clean_ref_file = clean_ref_file.loc[clean_ref_file['title'].str.len() >= min_chars]
+    clean_ref_file = clean_ref_file.drop_duplicates(subset = 'title')
+    #Converts to dict, for significantly faster run time over pandas DF
+    clean_ref_file = clean_ref_file.to_dict(orient = 'list')
```
nsorros: move the cleaning to the hard text search and pass the ref_file to it

utils/hard_text_search.py
```diff
@@ -0,0 +1,61 @@
...
+	string = string.lower()
+
+	return (string)
+
+
+def hard_text_search(scraped_text, ref_file, matches):
+	"""
+	Input:
+	-Raw scraped text (named tuple, SectionedDocument)
+	-Cleaned references file (dict, converted from pandas DF)
```
nsorros: this is a duplication, you pass both scraped text and clean references file, you should pass only one and clean if you need to inside this function

peter337:
Thanks @nsorros.
I think deduplication is already covered by line 46 of `hard_text_search.py`?

refparse.py
peter337: Wouldn't this slow the Tool down, as it would then be duplicating its cleaning of the UKRI pubs every time it looks at a new scraped document?

refparse.py
```diff
@@ -156,10 +168,16 @@ def run_predict(scraper_file, references_file,
             doc.uri
         )
 
-        matched_references = fuzzy_matcher.fuzzy_match(
+        fuzzy_matched_references = fuzzy_matcher.fuzzy_match(
```
lizgzil: I guess you called it this to distinguish between the algorithm matched data and the hard text search, but as nick say it's a bit outdated. I think `matched_references_parser` and  `matched_references_hard_text`.

Also I would do 
1. algorithm matches
2. hard text matches (but not for certain uber ids found in the algo matches)
3. merge the two match types

At the moment I guess the number 3 is in with number 2 which makes it all a bit less modular. I will make my suggestions on this on the relevant lines.

refparse.py
```diff
@@ -156,10 +168,16 @@ def run_predict(scraper_file, references_file,
...
             structured_references
         )
 
+        all_matched_references = hard_text_search(
```
lizgzil:
```
matched_references_parser["Tool"] = "Parser"
ref_exclude_list = matched_references_parser['WT_Ref_Id'].unique()
matched_references_hard_text = hard_text_search(doc, clean_ref_file, ref_exclude_list)
all_matched_references = pd.concat([matched_references_parser, matched_references_hard_text])
```
although it would be better to add ["Tool"] = "Parser" to the fuzzy match code not here (line 26).

utils/hard_text_search.py
```diff
@@ -0,0 +1,61 @@
...
+	-Extension to matches, with hard text search results concatenated
+	"""
+
+	#Adds a flag, showing which process found the match, and removes duplicates
+	matches['Tool'] = "Policy"
+	matches = matches.loc[:,~matches.columns.duplicated()]
+
+	clean_scraped_text = clean_text(scraped_text.section)
+
+	for i in range(len(ref_file['title'])):
```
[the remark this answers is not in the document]
peter337: Thanks Liz - unfortunately, I think this is required as `ref_file` is a dict, not a DF

peter337:
Hi @nsorros, @lizgzil, I've responded to your comments and I think it's ready for review again.

Few things to note:
-Character length limits (20 for Policy Tool, 40 for Hard Text search) are currently hard coded - may want to add them to settings.py?
-Since, for efficiency, `ref_file` is now a dict but is still being used like it's a DF, some of the stuff in `hard_text_search` looks a bit weird - looking for any suggestions to de-mess this piece of code.

I've also run a test on the msf publications, and it's now taking 1.5s per document. No idea where that increase has come from, but won't complain!

utils/hard_text_search.py
```diff
@@ -1,49 +1,61 @@
 import pandas as pd
 import re
 
+class HardTextSearch:
+	def __init__(self, ref_file, min_chars):
+		#Cleans up the ref file, changing from pandas DF to dict to reduce run time
+		ref_file['title'] = [self.clean_text(x) for x in ref_file['title']]
+		ref_file = ref_file.loc[ref_file['title'].str.len() >= min_chars]
+		ref_file = ref_file.drop_duplicates(subset = 'title')
+		ref_file = ref_file.to_dict(orient = 'list')
```
nsorros: all the above need to be stored in `self.ref_file`, i am not sure why or how the above works

peter337: I think it gets stored in `self.ref_file` on line 12?

refparse.py
```diff
@@ -168,16 +162,17 @@ def run_predict(scraper_file, references_file,
...
 
-        all_matched_references = hard_text_search(
+        matched_references_hard_text = text_matcher.hard_text_search(
             doc,
-            clean_ref_file,
-            fuzzy_matched_references
+            matched_references_parser['WT_Ref_Id'].unique()
         )
 
+        all_matched_references = pd.concat([matched_references_parser, matched_references_hard_text])
```
nsorros: that is not dealing with deduplication right?

refparse.py
```diff
@@ -126,6 +127,11 @@ def run_predict(scraper_file, references_file,
         settings.FUZZYMATCH_THRESHOLD
     )
 
+    text_matcher = HardTextSearch(
+        ref_file,
+        40
```
nsorros: add this number in settings

refparse.py
```diff
@@ -156,10 +162,17 @@ def run_predict(scraper_file, references_file,
...
         )
 
-        matched_references = fuzzy_matcher.fuzzy_match(
+        matched_references_parser = fuzzy_matcher.fuzzy_match(
             structured_references
         )
 
+        matched_references_hard_text = text_matcher.hard_text_search(
+            doc,
+            matched_references_parser['WT_Ref_Id'].unique()
```
nsorros: why are you passing the matched_references here? if for deduplication, this needs to happen in another function

refparse.py
```diff
@@ -170,7 +183,7 @@ def run_predict(scraper_file, references_file,
                     f"{doc.id}_{settings.PREF_REFS_FILENAME}"
                     )
                 )
-            matched_references.to_csv(
+            all_matched_references.to_csv(
```
nsorros: matched_references was a good name, no need to append with the all_ prefix

refparse.py
peter337: Deduplication should be being performed on lines 44-48 - do it before the hard text search in order to decrease run time.

utils/fuzzymatch.py
```diff
@@ -23,7 +23,8 @@ def match_vectorised(self, predicted_publications):
                 'Title': [],
                 'title': [],
                 'uber_id': [],
-                'Cosine_Similarity': []
+                'Cosine_Similarity': [],
+                'Tool': []
```
nsorros: I think the default here is firstly not a list but a string of some sort and also better to say "Parser" since even if it is empty it comes from here.

speaking of parser, parser does not feel like a good name since it is ambiguous. better call it fuzzy match and also tool is not helpful either as a column name, something like `Match algorithm` is better

utils/fuzzymatch.py
```diff
@@ -60,6 +61,11 @@ def match_vectorised(self, predicted_publications):
             }).reset_index()],
             axis=1)
 
+        match_data['Tool'] = "Parser"
+
+        match_data = match_data.loc[:,~match_data.columns.duplicated()]
+        match_data = match_data.loc[match_data['title'].str.len() >= 20]
```
nsorros: this is not a good place to do that, you should do that in the end after deduplication and outside this function and also the number should go to settings

utils/hard_text_search.py
```diff
@@ -0,0 +1,74 @@
+import pandas as pd
+import re
+
+class HardTextSearch:
+	def __init__(self, ref_file, min_chars):
+		#Cleans up the ref file, changing from pandas DF to dict to reduce run time
+		ref_file['title'] = [self.clean_text(x) for x in ref_file['title']]
```
nsorros: i don't think this changes from DF to dict. a better way to write what you are doing is ref_file['clean_title'] = ref_file['title'].apply(lambda x: self.clean_text(x))

it is also better to assign to a new column than overwrite

utils/hard_text_search.py
```diff
@@ -0,0 +1,74 @@
...
+		ref_file = ref_file.loc[ref_file['title'].str.len() >= min_chars]
+		ref_file = ref_file.drop_duplicates(subset = 'title')
```
nsorros: this has the implication that if two publications have the same title but are different publications for some reason one is chose randomly. at the moment we match only with title but in the future we might use additional info. if you keep this add a comment with this caveat.

utils/hard_text_search.py
```diff
@@ -0,0 +1,74 @@
...
+		ref_file = ref_file.to_dict(orient = 'list')
+
+		self.ref_file = ref_file
```
nsorros: ok i see, this is where the magic happens

utils/hard_text_search.py
```diff
@@ -0,0 +1,74 @@
...
+		Output:
+		-Extension to matches, with hard text search results concatenated
+		"""
+
+		#Removes refs already found by fuzzy matcher
+		indices_to_exclude = [i for i,x in enumerate(self.ref_file['uber_id']) if x in refs_to_exclude]
+		filtered_ref_file = self.ref_file
+		for i in sorted(indices_to_exclude, reverse = True):
+			del filtered_ref_file['title'][i]
+			del filtered_ref_file['uber_id'][i]
```
nsorros: do not use `del` and also if this is for deduplication it should go to another function possibly in refparse

utils/hard_text_search.py
peter337: Now that you've pointed it out, I don't think I will keep it.
I've remove the `subset='title'` argument, so now it's removing publications that share both a title and an ID. Now, if a unique publication is removed, it's an issue with the input data, not the code, as unique publications should not be sharing IDs.

refparse.py
nsorros: it should be done after, it is not about run time, it is about readability

utils/hard_text_search.py
nsorros: ok fair enough

utils/hard_text_search.py
```diff
@@ -0,0 +1,74 @@
...
+
+		clean_scraped_text = self.clean_text(scraped_text.section)
+
+		matches = pd.DataFrame()
+
+		for i in range(len(filtered_ref_file['title'])):
```
nsorros: you are iterating a dataframe inefficiently that is why you then need to do `filtered_ref_file['title'][i]`. Use iterrows so you iterate row by row and then you just need to select title and uber_id

utils/hard_text_search.py
```diff
@@ -0,0 +1,74 @@
...
+
+			title = filtered_ref_file['title'][i]
+			uber_id = filtered_ref_file['uber_id'][i]
+
+			#If there are new matches, append all relevant columns to the matched_refs DataFrame
+			if title in clean_scraped_text:
+
+				refs_matched_with_title = {
+					'Document id'       : scraped_text.id,
+					'Reference id'      : hash(title),
```
nsorros: that is a bit problematic and need to be discussed with the engineers as it might lead to duplicate data in the database.

utils/hard_text_search.py
peter337: It used to work like this, but now it's iterating over a dictionary as that was found to massively decrease the run time. Now that `filtered_ref_file` is a dict, I don't think iterrows will work?

utils/hard_text_search.py
```diff
@@ -0,0 +1,74 @@
...
+					'Title'             : title,
+					'WT_Ref_Title'      : title,
+					'WT_Ref_Id'         : uber_id,
+					'Cosine_Similarity' : 1,
```
nsorros: You do not need to fill all the fields. Since cosine similarity is not relevant here, you can just omit it

utils/hard_text_search.py
```diff
@@ -0,0 +1,74 @@
...
+			uber_id = filtered_ref_file['uber_id'][i]
+
+			#If there are new matches, append all relevant columns to the matched_refs DataFrame
+			if title in clean_scraped_text:
+
+				refs_matched_with_title = {
+					'Document id'       : scraped_text.id,
+					'Reference id'      : hash(title),
+					'Title'             : title,
+					'WT_Ref_Title'      : title,
```
nsorros: I think here you should put the unfiltered title

utils/hard_text_search.py
SamDepardieu: We're hashing the full text on our side, so this should be fine 👍

peter337: Just spoken to Sam - he's pointed out the fuzzy match hashes and done on the full text, while the text search hashes are done on just the title, so there should be no duplication.

peter337:
Hi @nsorros @lizgzil, just finished responding to the last review and ready for the next one.
Currently running tests to see how long the code takes per documents, will edit this comment once I have results.

utils/hard_text_search.py
```diff
@@ -0,0 +1,67 @@
...
+		-IDs of matches already found by fuzzy matcher (pandas Series)
+		Output:
+		-Extension to matches, with hard text search results concatenated
+		"""
+
+		clean_scraped_text = self.clean_text(scraped_text.section)
+
+		matches = pd.DataFrame()
+
+		for i in range(len(self.ref_file['clean_title'])):
```
nsorros: replace this with `for i, title in enumerate(self.ref_file['clean_title'])`

utils/hard_text_search.py
```diff
@@ -0,0 +1,67 @@
...
+					'Reference id'      : hash(title),
+					'Title'             : title,
+					'WT_Ref_Title'      : self.ref_file['title'][i],
+					'WT_Ref_Id'         : self.ref_file['uber_id'][i],
+					'Match_algorithm'   : "Text Search"
+				}
+
+				matches = matches.append(refs_matched_with_title, ignore_index=True)
+
+		return (matches)
```
nsorros: add a newline

settings.py
```diff
@@ -31,6 +31,9 @@ class BaseSettings:
 
     ORGANISATION_REGEX = "\\n[\d\.\s\\n]+(?=[A-Z])"
 
+    FUZZY_MATCH_CHAR_LIMIT = 20
```
nsorros: this is not for the fuzzy match, this is for all, maybe a more appropriate name is MIN_CHAR_LIMIT

settings.py
```diff
@@ -31,6 +31,9 @@ class BaseSettings:
...
+    HARD_TEXT_CHAR_LIMIT = 40
```
nsorros: Maybe name it HARD_TEXT_MAX_CHAR_LIMIT to avoid confusion because the other is called limit as well

refparse.py
```diff
@@ -94,6 +95,25 @@ def get_file(file_str, file_type, get_scraped=False):
     return file
 
 
+def remove_dups_and_concat(fuzzy_matches, text_matches):
+    #Remove duplicate columns and short matches in the fuzzy matches
+    fuzzy_matches = fuzzy_matches.loc[:,~fuzzy_matches.columns.duplicated()]
+    fuzzy_matches = fuzzy_matches.loc[fuzzy_matches['WT_Ref_Title'].str.len() >= settings.FUZZY_MATCH_CHAR_LIMIT]
+
+    if not text_matches.empty:
```
nsorros: that would be better in the beginning
```
if text_matches.empty:
    return fuzzy_matches
````

File: refparse.py [hunk 29]
```diff
@@ -94,6 +95,25 @@ def get_file(file_str, file_type, get_scraped=False):
...
+        duplicate_matches = fuzzy_matches['WT_Ref_Id'][fuzzy_matches['WT_Ref_Id'].isin(text_matches['WT_Ref_Id'])]
```
nsorros: these are not duplicate_matches but duplicate reference ids so maybe name it duplicate_ref_ids

peter337:
@nsorros All done :)

refparse.py
```diff
@@ -94,6 +95,25 @@ def get_file(file_str, file_type, get_scraped=False):
...
 
+def remove_dups_and_concat(fuzzy_matches, text_matches):
+    #Remove duplicate columns and short matches in the fuzzy matches
+    fuzzy_matches = fuzzy_matches.loc[:,~fuzzy_matches.columns.duplicated()]
+    fuzzy_matches = fuzzy_matches.loc[fuzzy_matches['WT_Ref_Title'].str.len() >= settings.MIN_CHAR_LIMIT]
+
+    if text_matches.empty:
+        return fuzzy_matches
+
+    else:
```
nsorros: you do not need the else

utils/hard_text_search.py
nsorros: It is better to hash("UBER_ID:"+str(uber_id))

refparse.py
```diff
@@ -94,6 +95,25 @@ def get_file(file_str, file_type, get_scraped=False):
...
+        duplicate_ref_ids = fuzzy_matches['WT_Ref_Id'][fuzzy_matches['WT_Ref_Id'].isin(text_matches['WT_Ref_Id'])]
```
lizgzil: No need to change as my method is more lines, but I was thinking how you could write this using the drop_duplicates function and I think this else can be written. I'm sure there must be a quicker way to to the update Match_algorithm bit!:
```
# Merge and deduplicate
merged_matches = pd.concat([text_matches, fuzzy_matches]).reset_index()
all_matches = merged_matches.drop_duplicates(subset = ['WT_Ref_Id'], keep = 'first')

# Update Match_algorithm
duplicated = merged_matches.duplicated(subset = ['WT_Ref_Id'], keep = False)
duplicate_indices = [i for i, d in enumerate(duplicated) if d and (i in all_matches.index)]
for i in duplicate_indices:
      all_matches.at[i, 'Match_algorithm'] = "Fuzzy and Text Searches"
```

refparse.py
```diff
@@ -94,6 +95,24 @@ def get_file(file_str, file_type, get_scraped=False):
...
+    fuzzy_matches = fuzzy_matches.loc[fuzzy_matches['WT_Ref_Title'].str.len() >= settings.MIN_CHAR_LIMIT]
+
+    if text_matches.empty:
+        return fuzzy_matches
+
+    duplicate_ref_ids = fuzzy_matches['WT_Ref_Id'][fuzzy_matches['WT_Ref_Id'].isin(text_matches['WT_Ref_Id'])]
+
+    #For duplicate matches: remove from text_matches, and renames 'Match_algorithm' in fuzzy_matches
+    text_matches = text_matches[~text_matches['WT_Ref_Id'].isin(duplicate_ref_ids)]
+    fuzzy_matches['Match_algorithm'][fuzzy_matches['WT_Ref_Id'].isin(duplicate_ref_ids)] = "Fuzzy and Text Searches"
```
lizgzil: This causes a warning:
```
A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy
  fuzzy_matches['Match_algorithm'][fuzzy_matches['WT_Ref_Id'].isin(duplicate_ref_ids)] = "Fuzzy and Text Searches"
./refparse.py:114: FutureWarning: Sorting because non-concatenation axis is not aligned. A future version
of pandas will change to not sort by default.

To accept the future behavior, pass 'sort=False'.

To retain the current behavior and silence the warning, pass 'sort=True'.
```
which I think you can fix by doing:
`fuzzy_matches.at[fuzzy_matches['WT_Ref_Id'].isin(duplicate_ref_ids), 'Match_algorithm'] = "Fuzzy and Text Searches"`

peter337: Nice catch Liz, thanks for the review :)

refparse.py
```diff
@@ -94,6 +95,26 @@ def get_file(file_str, file_type, get_scraped=False):
     return file
 
 
+def remove_dups_and_concat(fuzzy_matches, text_matches):
+    #Remove duplicate columns and short matches in the fuzzy matches
+    fuzzy_matches = fuzzy_matches.loc[:,~fuzzy_matches.columns.duplicated()]
```
nsorros: you can drop duplicates here instead

PR #102, opened 2019-03-11
add MIT license & update README for early open sourcing

hblanks:
# Description

As noted above, this adds a license and updates our README so it's more friendly to open source contributors and external users.

## Type of change

- [x] :memo: Documentation update

# How Has This Been Tested?

N/A

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,35 @@
 # Wellcome Policy Tool
 
-This repository will include all code needed to bring up Wellcome's
-Policy Tool, a service that reports which research publications are
-cited in the policy documents produced by policy organizations such as
-the WHO, MSF, and the UK government.
+Policy Tool is an open source service for discovering how research
+publications are being cited in global policy documents, including those
+produced by policy organizations such as the WHO, MSF, and the UK
+government. Key parts of it include:
+
+1. Web scrapers for pulling PDF documents from policy organizations,
+1. A reference parser for extracting references from these documents,
+1. A task for sourcing publications from EuropePMC,
+1. A task for matching references to publications, and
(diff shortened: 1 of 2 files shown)


hblanks:
(just chatted with @aoifespenge  - she'll take a look at the README tomorrow, but meantime we'll merge in the README update. But all future changes are welcome! Just moving the commits along. :-)

README.md
```diff
@@ -1,9 +1,35 @@
 # Wellcome Policy Tool
 
-This repository will include all code needed to bring up Wellcome's
-Policy Tool, a service that reports which research publications are
-cited in the policy documents produced by policy organizations such as
-the WHO, MSF, and the UK government.
+Policy Tool is an open source service for discovering how research
+publications are being cited in global policy documents, including those
+produced by policy organizations such as the WHO, MSF, and the UK
```
aoifespenge: Can you mention precisely which policy organisation data is currently available? I think it should be explicit here so that people have the right expectations for the tool

hblanks: Thanks for commenting! I confess, my thinking here was that we would be better off leaving the full list of organizations to some other part of the documentation, because prose tends to be harder to find and update than lists.

We will definitely have documentation reporting what organizations we have data from -- I suspect it may even be an online process, since users would ideally know when data was last scraped for every organization. Do you think it's OK for such detail to be provided either online or as a specific list elsewhere in the documentation, but not directly here in the prose?

aoifespenge: Yeah, that's fine by me if it is explicit elsewhere. Nicely written overall by the way.

PR #108, opened 2019-03-20
Fix parser failing

### LINKED ISSUE #105: reference parser fails with "Expected 2D array, got 1D array instead"
Sentry Issue: [REFERENCE-PARSER-8](https://sentry.io/organizations/wellcome-data-labs/issues/941250470/?project=1340876&referrer=github_integration)

```
Expected 2D array, got 1D array instead:
array=['World Health Organization' 'Global Tuberculosis Report 2016' 'WHO'
 'Geneva WHO/HTM/  TB/2016' '13'].
Reshape your data either using array.reshape(-1, 1) if your data has a single feature or array.reshape(1, -1) if it contains a single sample.
```

nsorros:
# Description

The problem was that when we switched to a pipeline pickle file consolidating the model and vectorizer together we did not update the settings and since the argument has remained the same for model file, it was causing a problem since the input was text to a model that was expecting numbers.

Fixes #105 

My only concern is that the way we run the parser in AWS also uses settings because if not this will not fix the problem and the problem needs to be fixed by updating the arguments there. @SamDepardieu @hblanks


SamDepardieu:
Short answer: from settings everywhere

Long answer:
At the moment we’re only running it with this:
 ```
env_vars = [
    "{\"name\": \"RDS_USERNAME\", \"value\": xxx"}",
    "{\"name\": \"RDS_PASSWORD\", \"value\": xxx"}",
    "{\"name\": \"RDS_HOST\", \"value\": \"xxx}\"}",
    "{\"name\": \"ORGANISATION\", \"value\": \"xxx\"}",
    "{\"name\": \"REF_PARSER_SETTINGS\", \"value\": \"xxx\"}",
    "{\"name\": \"SENTRY_DSN\", \"value\": \"xxx\"}",
  ]
```
on fargate and airflow’s task is not ready yet, but the dockerfile only runs:
`CMD ["python", "refparse.py"]`.
So as long as it’s still possible to run it like this with the up-mentioned env vars, everything will be fine.

SamDepardieu:
Btw you can also delete the old model files if we're not using them 👍

nsorros:
@SamDepardieu and @hblanks I deleted the old model files and I am merging this but in order for the bug to be fixed we need to deploy this. happy to wait for deployment to be done later or now but i might need your help to deploy it.

PR #112, opened 2019-04-09
Remove default location for references file

### LINKED ISSUE #81: Default location for Wellcome publications does not exist
Line 71 of 'settings.py' points towards a folder called 'wellcome_publications' as the default location to find publications if no location is supplied by the user. However, this folder does not exist.

nsorros:
Fixes #81 

This could be controversial so I am happy to be rejected or altered according to your suggestions. Peter encountered a problem when running the policy tool which I had in the past as well which is that the default location for references is non present for a user and is also wellcome centric.

My opinion is that there should be no default as the publications are an input that the user provides. There should maybe a sample but every user can place their references in a different location and that is fine. This is different from the output which to a certain extent we can dictate where is placed and is also different for the scraped files since these, to a certain extent, are produced by our code. The references on the other hand are being brough by the user.

An alternative I am thinking is to create an empty folder in the repo for the user to place their publications there.

Any other thoughts, suggestions do let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ mkdir -p ./tmp/parser-output/output_folder_name
 
 python ./refparse.py \
     --scraper-file "s3://datalabs-data/scraper-results/msf/20190117.json" \
-    --references-file "s3://datalabs-data/wellcome_publications/uber_api_publications.csv" \
+    --references-file "file://./references_folder/references_file.csv" \
     --model-file "s3://datalabs-data/reference_parser_models/reference_parser_pipeline.pkl" \
     --output-url "file://./tmp/parser-output/output_folder_name"
 ```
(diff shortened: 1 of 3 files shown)


lizgzil:
Good idea, in this case it might also be good to update the README with a more generic file location in the 'How to use it' section (like we did for `"file://./tmp/parser-output/output_folder_name"`).

nsorros:
@lizgzil you mean to change `s3://datalabs-data/wellcome_publications/uber_api_publications.csv`? and in particular the `wellcome_publications` part? in this case i think this is the link to our publications in s3, i am not expecting anyone to copy and paste. is it for stylistic reasons?

lizgzil:
> @lizgzil you mean to change `s3://datalabs-data/wellcome_publications/uber_api_publications.csv`? and in particular the `wellcome_publications` part? in this case i think this is the link to our publications in s3, i am not expecting anyone to copy and paste. is it for stylistic reasons?

yep that bit, but I think all the `wellcome_publications/uber_api_publications.csv` bit, not just `wellcome_publications`. e.g. change it to something like `file://./references_folder/references_file.csv`? it's just to guide the user to the fact they can enter their own file, but perhaps this isn't necessary!

nsorros:
I see, makes sense. Let me do that.

PR #127, opened 2019-04-18
Add stub for web UI

hblanks:
# Description

Build an extremely minimal web service for policytool using Falcon,
including one test. And make all other tests pass in policytool/.

Key points:

* Dockerfile:
  * add web/ to the image (we're going for a single image)
  * copy airflow.cfg & init.db into place for when we deploy it;
    this should have already been in.
* Makefile:
  * drop calver VERSION from docker images since we're building
    from codebuild
  * Install test requirements and run setup.py develop for development
    virtualenv.
  * Expand tests to run on both web/ and scraper/
* docker-compose.yml: add web container, update S3 path for airflow
* add initial webapp (web/api.py), templates, and blank CSS file
* Add web (and pdf_parser) to setup.py so they get installed
  by setup.py develop.
* Update README

## Type of change

- [x] :sparkles: New feature

# How Has This Been Tested?

```
make test
make docker-test
```

# Checklist:

- [] My code follows the style guidelines of this project (pep8 AND pyflakes) -- maybe? could we automate this in the Makefile?


policytool/Dockerfile
```diff
@@ -1,26 +1,25 @@
...
-
-# TODO: copy policytool/ or such once we're one dir up
 COPY setup.py /src/policytool/
 COPY README.md /src/policytool/
 COPY scraper/ /src/policytool/scraper/
 COPY pdf_parser/ /src/policytool/pdf_parser/
+COPY web/ /src/policytool/web/
 
+# No longer used?
 COPY ./entrypoint.sh /entrypoint.sh
```
SamDepardieu: Indeed, we should remove this. Can wait for another PR, though

policytool/docker-compose.yml
```diff
@@ -141,3 +141,23 @@ services:
       - worker
     volumes: *airflow-volumes
     depends_on: ["postgres", "redis"]
+
+  web:
+    # container_name: airflow-web
+    image: *airflow-image
```
SamDepardieu: I feel like this is a bit unclear, maybe we should change it to something like `policytool-image` 🤔 
Again nothing that needs to be done now

policytool/web/templates/index.html
```diff
@@ -0,0 +1,36 @@
...
+    }
+  </style> #}
+</head>
+<body><main>
+
+
+<section>
+  <h1>Wellcome Data Labs Policy Tool</h1>
+
+It's great.
```
SamDepardieu: True

policytool/docker-compose.yml
hblanks: yeah, the image is going to change name fairly soon. And probably start being pushed to docker hub...

policytool/Dockerfile
hblanks: Removed.

hblanks:
Update SQLAlchemy as well, fyi.

PR #128, opened 2019-04-24
Add a search tool to the web API

SamDepardieu:
# Description

This PR brings a search route to the API.

 - Define the route in api.py
 - Define the service in search.py
 - Change elastic/import_to_s3.py to also insert pdf_names and (randomly assigned for the time being) organisations
 - Add a persistent volume to the elasticsearch container

## Type of change
- [x] :sparkles: New feature

# How Has This Been Tested?

Mac OSx 10.14.4, Python3.6.4, Docker 18.09.5 with 2CPU/4Gb RAM

### DIFF
--- a/policytool/elastic/import_from_s3.py
+++ b/policytool/elastic/import_from_s3.py
@@ -5,6 +5,7 @@ import them into a running Elasticsearch database.
 import tempfile
 import csv
 import json
+import random
 
 import boto3
 from urllib.parse import urlparse
@@ -40,9 +41,15 @@ def write_to_es(es, line):
         line: a dict from a csv line. Should contain a file hash and the
               full text of a pdf
     """
+    # TODO: Use real pdf titles and orgs
+    # Select a random org for the time being
+    orgs = ['who', 'nice', 'msf']
+    org = random.choice(orgs)
     body = json.dumps({
         'hash': line['file_hash'],
-        'text': line['pdf_text']
+        'text': line['pdf_text'],
+        'title': line['pdf_name'],
+        'organisation': org,
     })
     es.index(
         index='datalabs-fulltexts',
--- a/policytool/web/api.py
+++ b/policytool/web/api.py
@@ -8,14 +8,13 @@ Entrypoint for web UI & API. Run with:
 import logging
 import os
 import os.path
+from urllib.parse import urlparse
 
 import falcon
 import jinja2
-
-from urllib.parse import urlparse
 from elasticsearch import Elasticsearch
 
-from web import search
+from . import search
 
 TEMPLATE_ROOT = os.path.join(os.path.dirname(__file__), 'templates')
 STATIC_ROOT = os.environ.get('STATIC_ROOT')
@@ -117,18 +116,14 @@ configure_logger(logger)
 assert os.environ.get('ELASTICSEARCH_HOST'), 'No elasticsearch host'
 parsed_url = urlparse(os.environ['ELASTICSEARCH_HOST'])
 
-try:
-    logger.info('Trying to connect to {elastic_host}'.format(
-        elastic_host=os.environ['ELASTICSEARCH_HOST']
-    ))
-    es = Elasticsearch([{
-            'host': parsed_url.hostname,
-            'port': parsed_url.port
-        }],
-    )
-except Exception as e:
-    raise e
-
+logger.info('Connecting to {elastic_host}'.format(
+    elastic_host=os.environ['ELASTICSEARCH_HOST']
+))
+es = Elasticsearch([{
+        'host': parsed_url.hostname,
+        'port': parsed_url.port
+    }],
+)
 
 # Routes (are LIFO)
 api = falcon.API()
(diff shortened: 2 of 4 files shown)


SamDepardieu:
[Allow user to search a policy document by a REST API](https://trello.com/c/fUzpcFz1/670-allow-user-to-search-a-policy-document-by-a-rest-api)

policytool/web/search.py
```diff
@@ -0,0 +1,86 @@
...
+            else:
+                resp.body = json.dumps({
+                    'status': 'error',
+                    'message': response
+                })
+        else:
+            resp.body = json.dumps({
+                'status': 'error'
+            })
+            resp.status = falcon.HTTP_200
```
hblanks: Seems like this is an invalid request, so not 200?

policytool/web/api.py
```diff
@@ -104,11 +112,31 @@ def on_get(self, req, resp):
...
+    logger.info('Trying to connect to {elastic_host}'.format(
+        elastic_host=os.environ['ELASTICSEARCH_HOST']
+    ))
+    es = Elasticsearch([{
+            'host': parsed_url.hostname,
+            'port': parsed_url.port
+        }],
+    )
+except Exception as e:
+    raise e
```
hblanks: Like Master Yoda would say, "There is no try" here, just do or do not :-)

policytool/web/api.py
```diff
@@ -12,6 +12,12 @@
 import falcon
 import jinja2
 
+from urllib.parse import urlparse
+
+from elasticsearch import Elasticsearch
+
+from web import search
```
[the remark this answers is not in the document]
SamDepardieu: You can only use relative import inside a `from x import y` if I remember well, but I sure could use a `from . import search` 👍

policytool/web/search.py
SamDepardieu: Whoopsy

policytool/web/search.py
```diff
@@ -0,0 +1,86 @@
...
+            return True, self.es.search(
+                index='datalabs-fulltexts',
+                body=json.dumps(es_body),
+                explain=True
+            )
+
+        except ConnectionError:
+            message = 'Could not join the elasticsearch server'
+            api.logger.error(message)
+            return False, message
```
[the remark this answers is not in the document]
SamDepardieu: If I raise it here, it makes the server crash when elasticsearch isn't ready yet to accept connections. The idea between the return was to fail without getting a 500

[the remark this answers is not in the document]
SamDepardieu: We should still set the status to a 50x (504 I think?) though

policytool/web/api.py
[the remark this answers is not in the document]
hblanks: Either turns out to work -- search for "import ." in https://www.python.org/dev/peps/pep-0328/ for reference.

But, imports like this do only work when you're inside a "package." This is complicated but I can explain if you like some time.

policytool/web/search.py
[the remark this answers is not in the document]
hblanks: Interesting! Does the server crash or just the handler for that HTTP request? I'd just taken exception handling for granted...

Docs say use falcon's HTTPError or some variant with the right status -- that sounds right to me. Cf. https://falcon.readthedocs.io/en/stable/api/errors.html . In that case the error logging is also very well placed; we'd rely on that for Sentry reporting.

I'd return a 503 b/c your dependency is not exactly an HTTP backend.

policytool/elastic/import_from_s3.py
```diff
@@ -40,9 +41,15 @@ def write_to_es(es, line):
         line: a dict from a csv line. Should contain a file hash and the
               full text of a pdf
     """
+    # TODO: Use real pdf titles and orgs
+    # Select a random org for the time being
+    orgs = ['who', 'nice', 'msf']
+    org = random.choice(orgs)
```
nsorros: what's going on here? :p

policytool/web/search.py
```diff
@@ -0,0 +1,88 @@
...
+            req_params: The request's parameters. Shoud include 'term' and at
+                        least a field (on of [text|title|organisation])
+
+        Returns:
+            True|False: The search success status
+            es.search()|str: A dict containing the result of the search if it
+                             succeeded or a string explaining why it failed
+        """
+        try:
+            fields = req_params.get('fields', [])
```
nsorros: maybe check that fields elements are valid

SamDepardieu: That's no issue if they are not, it'll just find no matches

policytool/elastic/import_from_s3.py
SamDepardieu: I don't have the real data in the dataset I'm unsing here, so I give it a random placeholder instead :D

policytool/elastic/import_from_s3.py
```diff
@@ -5,6 +5,7 @@
 import tempfile
 import csv
 import json
+import random
```
hblanks: @nsorros  - You asked a great question later in this file. To go one higher in your tech lead skills, just ask your reviewee

> what's going on here?

not at invocation time but at import.

PR #133, opened 2019-04-29
Fix tests not passing because of missing ES host

SamDepardieu:


hblanks:
Maybe we set up a wsgi.py that import os.environ and remove deps on os.environ from api.py?

Per https://falcon.readthedocs.io/en/stable/deploy/nginx-uwsgi.html

PR #134, opened 2019-04-30
Neaten algorithm tests

lizgzil:
**At the moment the structure of this folder is what needs reviewing, after this the functions will be filled in**

# Description

This PR adds `test_algo.py` and the 'algo_tests' folder. These files will allow the user to test the different parts of the algorithm.

There are 7 tests in total, although one (test 5) is optional (should we print here that the accuracy of the reference parser can be found elsewhere, or should we test the model again here?).

The data needed is [here](https://s3-eu-west-1.amazonaws.com/datalabs-data/policy_tool_tests), but in the future we might want to put this elsewhere.

## To do
- [ ] check with @SamDepardieu how to complete the scrape test code (in `test_scrape.py`)
- [ ] delete everything in the old `model_tests` folder
- [ ] complete all the `test_metric` functions
- [ ] attractively save to file and/or print to terminal all the results
- [ ] get the test data for several of the tests (this is needed for `scrape_test_data.csv`, `split_reference_test_data.csv`, `merge_components_test_data.csv`, which currently have made up data but in the correct structure)
- [ ] relocate the test data

## Type of change

- [x] :sparkles: New feature
- [ ] :fire: Breaking change

# How Has This Been Tested?

It hasn't.

# Checklist:


algo_tests/test_scrape.py
```diff
@@ -0,0 +1,47 @@
...
+    metric = 50 # Place holder
+
+    return metric
+
+def scrape_urls(scrape_test_data):
+
+    references_text = []
+
+    for url in scrape_test_data['url']:
+        # data = scrape_function(url) #### NO IDEA HERE - SCRAPE THE URL GIVEN
```
nsorros: We do not want to test or depend on how well the scraper, scrapes pdfs from URLs. Our starting point is the pdfs or the text from the pdfs here. We want to test how well we extract the references section from a pdf (or the text of the pdf).

algo_tests/test_scrape.py
```diff
@@ -0,0 +1,47 @@
...
+        else:
+            references_text.append(None) # It might be that nothing is scraped
+
+    scrape_test_data['References section text scraped'] = references_text
+
+    return scrape_test_data
+
+def test_scrape(scrape_test_data):
+
+    test1_2_info = scrape_urls(scrape_test_data)
```
nsorros: continuing from the comment above, to do so we need a function that takes a pdf and returns a reference section if there is one. this function probably lives somewhere in sams code and it can be a rule based approach or a machine learning model soon but in any case this function should provide us now and in the future with the reference section given a pdf or the raw text of a pdf.

the data in this case are on the one hand a folder of pdfs (or files containing the full text of the pdfs) and on the other hand at the very least a file saying whether a particular pdf has a references section.

algo_tests/test_split_reference.py
```diff
@@ -0,0 +1,28 @@
...
+from utils.parse import split_reference
+
+
+def test_metric(actual, predicted):
+
+    metric = 50 # Place holder
+
+    return metric
+
+def test_split_reference(split_reference_test_data):
```
nsorros: I am not sure how we will test this. I understand this is part of the parse step which is a three-step process of 1) split to components 2) predict components 3) merge components but at the same time there is no right components to split to, are there? It might be better at this point to test the entire parse function which takes a reference as an input and return a structured reference than the individual components and write, at a later time, software tests for the individual components to make sure the functionality works as expected. Some tests exist already.

algo_tests/test_split_section.py
```diff
@@ -0,0 +1,31 @@
+import pandas as pd
+from utils.split import split_section
+
+def test_metric(actual, predicted):
+
+    metric = 50 # Place holder
+
+    return metric
+
+def test_split_section(split_section_test_data):
```
nsorros: In this test, I think we should do what we have been doing when developing the ML solution. This is to have one file per section in a folder and then one file that contains the information about how many references a section has. I think this is what you have done, let me know if not.

algo_tests/test_merge_components.py
```diff
@@ -0,0 +1,33 @@
...
+
+from utils.parse import merge_components
+
+def test_metric(actual, predicted):
+
+    metric = 50 # Place holder
+
+    return metric
+
+def test_merge_components(merge_components_test_data):
```
nsorros: Similar to split_reference, I think this is a difficult test to define. The output of merge components is relatively objective but what about the input? In particular, what are the right components to expect there? Single words, more than single words? It depends how we split. I think the easiest thing to test is the entire parse function. To a certain extent merge components can always be coded to work perfectly given ideal inputs.

algo_tests/test_match_vectorised.py
```diff
@@ -0,0 +1,40 @@
...
+def test_metric(test_info):
+    """
+    False positive rate or similar
+    """
+
+    metric = 50 # Place holder
+
+    return metric
+
+def test_match_vectorised(match_publications, test_publications, match_threshold):
```
nsorros: In this test to reiterate on our conversation, we would like a set of titles that match in a given big set of publications (wellcome publications), as well as an equal amount of set of titles that do not match. The metric will then be an accuracy type of metric.

In case we need to reconstruct the sample dataset, the way we went on about it, which should be documented somewhere, is that we picked 100 titles that had matched and found the 1st and 2nd match from the list of publications and then the first results would create the list of titles that match while the second results would create the list that do not match. We then went and validated that list to make sure that indeed our classification of the first as matching and second as not matching is correct

nsorros:
On a more subtle note, I think you should avoid using words like neaten, clean, temp etc in PR names, descriptions and code. In particular, for this PR instead of naming it Neaten algorithm tests which is ambiguous, I would go for Fix, Update or Rewrite which is slightly more clear and more often used.

Similarly, test_algo and algo_tests is also ambigious since the entire code of the policy tools contains algorithms. I think at the very least you should replace algo with ml (machine learning) but ideally you should move away from using the word tests as well since tests point to software engineering tests which is not what we are doing. I would go for evaluation and evaluate which i think differentiates from the overloaded test word and I feel better describes what we are doing.

lizgzil:
> Similarly, test_algo and algo_tests is also ambigious since the entire code of the policy tools contains algorithms. I think at the very least you should replace algo with ml (machine learning) but ideally you should move away from using the word tests as well since tests point to software engineering tests which is not what we are doing. I would go for evaluation and evaluate which i think differentiates from the overloaded test word and I feel better describes what we are doing.

I will change to evaluate, but I didn't want to say 'ml' or 'models' since the point of these tests is indeed to test different parts of the entire policy tool (not just the ml or model bits), which is why I chose 'test_algo'. Maybe evaluate_policy_tool then ? Although slightly less appealing with the name of the policy tool changing soon.

algo_tests/test_split_section.py
lizgzil: yep, I was planning on using the same data

algo_tests/test_scrape.py
lizgzil: 👍 I've changed 'url' to 'pdf', but I think this code will eventually change anyway once we know exactly what the input data will look like (mostly if we choose to have a folder of pdfs/their text).

algo_tests/test_split_reference.py
lizgzil: i've now merged these 3 parts into 1 test (test_parse) which will test `structure_reference` from `parse.py`

algo_tests/test_merge_components.py
lizgzil: the merging of this into `test_parse` deals with this issue

algo_tests/test_match_vectorised.py
lizgzil: i've added `evaluation_data.md` which is the place I will discuss how each evaluation data was created

algo_evaluation/evaluate_settings.py
```diff
@@ -0,0 +1,22 @@
+import os
+from settings import BaseSettings
+
+
+class TestSettings(BaseSettings):
```
nsorros: why you inherit from BaseSettings?

evaluate_algo.py
```diff
@@ -0,0 +1,135 @@
...
+    logger.info('[+] Running test 4')
+    test4_info, test4_score = evaluate_parse(parse_test_data, model)
+
+    logger.info('[+] Running test 5')
+    test5_info, test5_score = evaluate_match_vectorised(match_publications, test_publications, settings.FUZZYMATCH_THRESHOLD)
+
+
+    log_file.write('=====\nTest 1 results:\n=====\n')
+    if args.verbose == 'True':
+        print("Lots of information about test 1")
```
nsorros: either log or print

algo_evaluation/evaluate_settings.py
```diff
@@ -0,0 +1,22 @@
...
+    DEBUG = True
```
nsorros: why do you need a debug?

algo_evaluation/evaluate_parse.py
```diff
@@ -0,0 +1,25 @@
...
+    for reference in parse_test_data['Actual reference']:
+
+        structured_reference = structure_reference(model, reference)
+
+        predicted_structure.append(structured_reference)
+
+    parse_test_data['Predicted merged components'] = predicted_structure
+
+    test_info = parse_test_data
+    test_score = evaluate_metric(parse_test_data['Title'], parse_test_data['Predicted merged components'])
```
nsorros: here you pass title and predicted components which is not like for like, is it? i would suggest to pass the information about all fields in both actual and predict case

algo_evaluation/evaluate_match_vectorised.py
```diff
@@ -0,0 +1,40 @@
...
+
+    match_data = fuzzy_matcher.match_vectorised(test_publications)
+
+    test_info = pd.merge(
+        test_publications[['Reference id', 'Type', 'Title', 'Document id']],
+        match_data[['Reference id', 'uber_id', 'Cosine_Similarity', 'title']],
+        on='Reference id', how='left'
+        )
+
+    test_score = evaluate_metric(test_info)
```
nsorros: here we need to pass two vectors, one indicates whether the publications should be matched and the other whether it was matched. the first vector exists in the datasets, the second should be constructed, for example by running the publications one by one to the matchers and recording 1 or 0 if there is a match or not (if you can think of a vectorised implementation it is welcome)

algo_evaluation/evaluate_settings.py
lizgzil: it's to get the other variables from settings that I need (ORGANISATION_REGEX, FUZZYMATCH_THRESHOLD)

algo_evaluation/evaluate_settings.py
lizgzil: taken out

algo_evaluation/evaluate_settings.py
nsorros: 👍

algo_evaluation/evaluate_parse.py
lizgzil: 👍

lizgzil:
@nsorros changes made (all your required ones).

algo_evaluation/evaluate_find_section.py
```diff
@@ -0,0 +1,47 @@
...
+
+def evaluate_metric_quality(actual, predicted):
+    """
+    Text similarity
+    """
+    metric = 50 # Place holder
+
+    return metric
+
+def scrape_urls(scrape_test_data):
```
nsorros: this should probably be called scrape_pdfs

README.md
```diff
@@ -103,5 +103,13 @@ You can run the unittests for this project by running:
 
 This will test that your last changes didn't affect how the program works.
 
+## Testing each component of the algorithm
+
+We have devised some test data in order to test 7 steps of the model. The results can be found by first downloading the test data from [here](https://s3-eu-west-1.amazonaws.com/datalabs-data/policy_tool_tests), and then running
```
nsorros: I think the tests are 5 now

PR #144, opened 2019-05-10
Add evaluate split section metric

lizgzil:
Add metric calculation for evaluate split section and make the print out in a nice format

Note that the new functions are largely copied and pasted from [here](https://github.com/wellcometrust/datalabs/blob/bced9699254199ecff76edeca1ce80cefbaa9a01/modelling/policy/references-splitting/src/evaluate_splitter.py) (where `calc_num_metric` and `evaluate_splitter_nums` are practically the same - I return an extra variable in the latter)

# Description

This PR adds some functions to `evaluate_split_section.py` to calculate this evaluation metric.
I also edit the log file print in `evaluate_algo.py` so that it prints the information in a nice format, which now looks like:

```
-----Information about test 2:-----

Score
52.8

Percentage below threshold of 40
52.8

Median difference
29.6

Metrics grouped by source
          Median difference  Percentage below threshold of 40
Source                                                       
msf                     7.3                              83.3
nice                   95.8                               0.0
unicef                 27.4                              90.0
who_iris               10.0                              82.4
```

## Type of change

- [x] :sparkles: New feature
- [ ] :fire: Breaking change

# How Has This Been Tested?

No tests.

# Checklist:

### DIFF
--- a/policytool/refparse/algo_evaluation/evaluate_split_section.py
+++ b/policytool/refparse/algo_evaluation/evaluate_split_section.py
@@ -34,30 +34,31 @@ def calc_bel_metric(metrics, threshold):
 
     return bel_metric
 
-def evaluate_metric(evaluation_info, threshold):
+def evaluate_metrics(actual_pred_num_references, threshold):
 
 
-    evaluation_info['diff_metric'] = [
-                                        calc_num_metric(m,n) for m,n in zip(
-                                            evaluation_info["Predicted number of references"],
-                                            evaluation_info["Number of references scraped"]
-                                            )
-                                    ]
-    median_diff = calc_med_metric(evaluation_info['diff_metric'])
-    below_threshold = calc_bel_metric(evaluation_info['diff_metric'], threshold)
-    grouped_source_metrics = evaluation_info.groupby('Source')['diff_metric'].agg(
-        {'Median difference': lambda x: calc_med_metric(x),
-        'Percentage below threshold of {}'.format(threshold) : lambda x : calc_bel_metric(x, threshold)}
+    actual_pred_num_references['diff_metric'] = [
+        calc_num_metric(m,n) for m,n in zip(
+            actual_pred_num_references["Predicted number of references"],
+            actual_pred_num_references["Number of references scraped"]
         )
+    ]
 
-    test_scores = {
+    median_diff = calc_med_metric(actual_pred_num_references['diff_metric'])
+    below_threshold = calc_bel_metric(actual_pred_num_references['diff_metric'], threshold)
+    grouped_source_metrics = actual_pred_num_references.groupby('Source')['diff_metric'].agg({
+        'Median difference': lambda x: calc_med_metric(x),
+        'Percentage below threshold of {}'.format(threshold) : lambda x : calc_bel_metric(x, threshold)
+    })
+
+    metrics = {
         'Score' : below_threshold,
         'Percentage below threshold of {}'.format(threshold) : below_threshold,
         'Median difference' : median_diff,
         'Metrics grouped by source' : grouped_source_metrics
         }
 
-    return test_scores
+    return metrics
 
 def evaluate_split_section(split_section_test_data, regex, threshold):
     """
@@ -65,17 +66,15 @@ def evaluate_split_section(split_section_test_data, regex, threshold):
     and compare the findings with the ground truth
     """
 
-    evaluation_info = []
+    actual_pred_num_references = []
     for references_section in split_section_test_data['Reference section']:
         references = split_section(references_section, regex = regex)
-        evaluation_info.append(
-            {
+        actual_pred_num_references.append({
             "Predicted references" : references,
             "Predicted number of references" : len(references)
-            }
-            )
-    evaluation_info = pd.concat([split_section_test_data, pd.DataFrame(evaluation_info)], axis = 1)
+        })
+    actual_pred_num_references = pd.concat([split_section_test_data, pd.DataFrame(actual_pred_num_references)], axis = 1)
 
-    test_scores = evaluate_metric(evaluation_info, threshold)
+    metrics = evaluate_metrics(actual_pred_num_references, threshold)
 
-    return test_scores
+    return metrics
--- a/policytool/refparse/algo_evaluation/evaluate_split_section.py
+++ b/policytool/refparse/algo_evaluation/evaluate_split_section.py
@@ -19,43 +19,36 @@ def calc_num_metric(predicted_number, actual_number):
 
     return metric
 
-def evaluate_splitter_nums(number_predicted_references, number_actual_references, threshold):
-    """
-    Calculates the number of references test metric for each document.
-    Input:
-        number_predicted_references : A list of the predicted numbers
-                                        of references from a sample of documents
-        number_actual_references : A list of the actual numbers
-                                        of references from a sample of documents
-        threshold : a threshold for the percentage difference acceptable to be below
-    Output:
-        median_diff : the median percentage difference for a sample of documents
-        below_threshold : the percentage of documents with a test metric less than the threshold
-    """
-   
-    doc_metrics = [calc_num_metric(m,n) for m,n in zip(number_predicted_references, number_actual_references)]
+def calc_med_metric(metrics):
 
-    median_diff = round(np.median(doc_metrics), 1)
-    below_threshold = round(
-        100 * len([x for x in doc_metrics if x <= threshold]) / len(doc_metrics),
+    med_metric = round(np.median(metrics), 1)
+
+    return med_metric
+
+def calc_bel_metric(metrics, threshold):
+
+    bel_metric = round(
+        100 * len([x for x in metrics if x <= threshold]) / len(metrics),
         1
     )
 
-    return doc_metrics, median_diff, below_threshold
+    return bel_metric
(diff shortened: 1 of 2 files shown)


algo_evaluation/evaluate_split_section.py
```diff
@@ -1,9 +1,68 @@
...
+    as a percentage of the actual number
+    """
+
+    # In the case of 0 references (some documents don't have references but do
+    # have a reference section), set actual_number to the closest integer, 1.
+    # This introduces some small error, but avoids the script throwing a math error
+    if actual_number == 0:
+        actual_number = 1
+
+    metric = abs(100*((predicted_number - actual_number) / actual_number))
```
nsorros: didn't we change that metric with Peter?

algo_evaluation/evaluate_split_section.py
```diff
@@ -1,9 +1,68 @@
...
+    """
+    Calculates the number of references test metric for each document.
+    Input:
+        number_predicted_references : A list of the predicted numbers
+                                        of references from a sample of documents
+        number_actual_references : A list of the actual numbers
+                                        of references from a sample of documents
+        threshold : a threshold for the percentage difference acceptable to be below
+    Output:
+        median_diff : the median percentage difference for a sample of documents
```
nsorros: you also output doc_metrics

algo_evaluation/evaluate_split_section.py
nsorros: maybe not actually, i was thinking the median which you do later

algo_evaluation/evaluate_split_section.py
```diff
@@ -1,9 +1,68 @@
...
+
+    source_metric = pd.DataFrame(
+        [list(sources), doc_metrics]
+        ).transpose().rename(columns = {0 : "Source", 1 : "Metric"})
+
+    aggregations = {
+            'Median difference' : lambda x : round(np.median(x), 1),
+            'Percentage below threshold of {}'.format(threshold) : lambda x : round(100 * len([y for y in x if y <= threshold]) / len(x),1)
+    }
+    grouped_source_metrics = source_metric.groupby('Source')['Metric'].agg(aggregations)
```
nsorros: You are repeating code here. It would be better to do the calculation per source instead of returning the doc_metrics and rewriting the same code. 

The best way to do that is to create the following functions, some of which exist already
* calc_num_metric: calculate difference between predict and actual given predicted and actual
* calc_med_metric: calculate median given an array
* calc_bel_metric: calculate below threshold percentage given an arrray

Then in your name where you have the data in the format source, predicted, actual you do the following
```
data['diff_metric'] = calc_num_metric(...)
data['med_metric'] = data['diff_metric'].apply(calc_med_metric)
data['below_metric'] = data['diff_metric'].apply(calc_bel_metric)
data['group_metrics'] = data.groupby('source').agg({'Med dff': lambda x: calc_med_metric}, ...)
```

This is very close to what you have but it ensures that if you change the definition of how to calculate median or below threshold metric you do not need to change it in two places.

lizgzil: i've made these changes, but in a slightly different way to how you suggested - I left them in the evaluate_metric function.

lizgzil:
@nsorros I've made the changes like you said

policytool/refparse/evaluate_algo.py
```diff
@@ -30,7 +30,7 @@ def create_argparser():
     parser.add_argument(
         '--verbose',
         help='Whether you want to print detailed test information ("True") or not ("False")',
-        default = False
+        default = True
```
nsorros: bad liz

policytool/refparse/algo_evaluation/evaluate_split_section.py
```diff
@@ -1,9 +1,61 @@
...
+    return metric
+
+def calc_med_metric(metrics):
+
+    med_metric = round(np.median(metrics), 1)
+
+    return med_metric
 
-    test_scores = {'Score' : 50, 'More info' : "More information about this test"}
+def calc_bel_metric(metrics, threshold):
```
nsorros: cal_bel_metric for some reason expands to calc_belly_metric in my mind :p

policytool/refparse/algo_evaluation/evaluate_split_section.py
```diff
@@ -1,9 +1,61 @@
...
+        100 * len([x for x in metrics if x <= threshold]) / len(metrics),
+        1
+    )
+
+    return bel_metric
+
+def evaluate_metric(evaluation_info, threshold):
+
+
+    evaluation_info['diff_metric'] = [
```
nsorros: formating comment:
```
evaluation_info['diff_metric'] = [
    calc_num_metric(m,n) for m,n in zip(
        evaluation_info["Predicted number of references"],
       evaluation_info["Number of references scraped"]
       )
]
```

policytool/refparse/algo_evaluation/evaluate_split_section.py
```diff
@@ -1,9 +1,61 @@
...
+                                        calc_num_metric(m,n) for m,n in zip(
+                                            evaluation_info["Predicted number of references"],
+                                            evaluation_info["Number of references scraped"]
+                                            )
+                                    ]
+    median_diff = calc_med_metric(evaluation_info['diff_metric'])
+    below_threshold = calc_bel_metric(evaluation_info['diff_metric'], threshold)
+    grouped_source_metrics = evaluation_info.groupby('Source')['diff_metric'].agg(
```
nsorros: formatting comment:
```
grouped_source_metrics = evaluation_info.groupby('Source')['diff_metric'].agg({
    'Median difference': lambda x: calc_med_metric(x),
    'Percentage below threshold of {}'.format(threshold) : lambda x : calc_bel_metric(x, threshold)
})
```

policytool/refparse/algo_evaluation/evaluate_split_section.py
```diff
@@ -13,19 +65,17 @@ def evaluate_split_section(split_section_test_data, regex, threshold):
     and compare the findings with the ground truth
     """
 
-    comparison = []
+    evaluation_info = []
     for references_section in split_section_test_data['Reference section']:
         references = split_section(references_section, regex = regex)
-
```
nsorros: it would be nice if this code worked without the need for regex since when you switch to an approach that does not require regex it might break

policytool/refparse/algo_evaluation/evaluate_split_section.py
```diff
@@ -13,19 +65,17 @@ def evaluate_split_section(split_section_test_data, regex, threshold):
     and compare the findings with the ground truth
     """
 
-    comparison = []
+    evaluation_info = []
```
nsorros: evaluation_info is not a good name

policytool/refparse/algo_evaluation/evaluate_split_section.py
```diff
@@ -13,19 +65,17 @@ def evaluate_split_section(split_section_test_data, regex, threshold):
...
             "Predicted references" : references,
             "Predicted number of references" : len(references)
             }
             )
-    comparison = pd.concat([split_section_test_data, pd.DataFrame(comparison)], axis = 1)
+    evaluation_info = pd.concat([split_section_test_data, pd.DataFrame(evaluation_info)], axis = 1)
 
-    test_info = comparison
-    test_scores = evaluate_metric(comparison["Number of references scraped"], comparison["Predicted number of references"])
+    test_scores = evaluate_metric(evaluation_info, threshold)
```
nsorros: scores or metrics would work as well as a name

policytool/refparse/algo_evaluation/evaluate_split_section.py
```diff
@@ -13,19 +65,17 @@ def evaluate_split_section(split_section_test_data, regex, threshold):
     and compare the findings with the ground truth
     """
 
-    comparison = []
+    evaluation_info = []
     for references_section in split_section_test_data['Reference section']:
         references = split_section(references_section, regex = regex)
-
-        comparison.append(
+        evaluation_info.append(
```
nsorros: the formatting below can be improved

policytool/refparse/algo_evaluation/evaluate_split_section.py
lizgzil: you can apply this function to your waist to calculate your trouser size

policytool/refparse/algo_evaluation/evaluate_split_section.py
lizgzil: hm, well at the moment we have `split_section(references_section, regex="\n")` in `split.py`, so I guess the problem here could be solved if we changed the default regex to the universal regex not "\n". Shall I do this in this PR or a new one? It does seem to me like we should have the default as the universal regex anyway.

policytool/refparse/algo_evaluation/evaluate_split_section.py
lizgzil: changed to 'actual_pred_num_references'

policytool/refparse/algo_evaluation/evaluate_split_section.py
lizgzil: for the sake of closing this PR, I will make these changes in a new PR and write an issue for the interim

PR #148, opened 2019-05-15
Add evaluate parser metric

lizgzil:
# Description

Adding the metric function in `evaluate_parse.py`. This now returns a number of metrics:
- Proportion of components predicted correctly (exact string matches)
- Mean normalised Levenshtein distance (the smaller the better)
- Proportion of components predicted almost correctly (Levenshtein < threshold)

I return these metrics for all (macro) and by each category (micro).

I also add the levenshtein threshold parameter to `evaluate_settings` and `evaluate_algo`. And I also needed to add the editdistance import to the unpinned requirements file - @hblanks do I need to do anything else concerning adding this new import?

## Type of change

- [ ] :sparkles: New feature

# How Has This Been Tested?

No tests.

# Checklist:


nsorros:
I think you need to run `make update-requirements-txt` and commit the requirements file as well. @hblanks can confirm.

nsorros:
When running `evaluate_algo.py` I encountered this problem. Is that intended?

```
[2019-05-16 10:53:49,850]:INFO - [+] Running tests 1 and 2
Traceback (most recent call last):
  File "/Users/sorrosn/.virtualenvs/reference_parser/lib/python3.7/site-packages/pandas/core/indexes/base.py", line 3078, in get_loc
    return self._engine.get_loc(key)
  File "pandas/_libs/index.pyx", line 140, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/index.pyx", line 162, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 1492, in pandas._libs.hashtable.PyObjectHashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 1500, in pandas._libs.hashtable.PyObjectHashTable.get_item
KeyError: 'pdf'```

File: policytool/refparse/algo_evaluation/evaluate_parse.py [hunk 1]
```diff
@@ -1,34 +1,87 @@
...
-    'Volume': parse_test_data['Volume'],
-    'Issue': parse_test_data['Issue'],
-    'Pagination': parse_test_data['Pagination'],
-    'Title': parse_test_data['Title'],
-    'PubYear': parse_test_data['PubYear']
-    }
-
-    test_info = parse_test_data
-    test_scores = evaluate_metric(merged_components, parse_test_data['Predicted merged components'])
+    component_names = ['Authors', 'Journal', 'Volume', 'Issue', 'Pagination', 'Title', 'PubYear']
```
nsorros: these are not components. components are the splitted strings from the reference. these are the category names. rename.

policytool/refparse/algo_evaluation/evaluate_parse.py
```diff
@@ -1,34 +1,87 @@
...
+def evaluate_parse(parse_test_data, model, levenshtein_threshold):
 
     predicted_structure = []
     for reference in parse_test_data['Actual reference']:
-
         structured_reference = structure_reference(model, reference)
-
         predicted_structure.append(structured_reference)
 
     parse_test_data['Predicted merged components'] = predicted_structure
```
nsorros: why do you need to save that in the dataframe and then transform it into a list later. drop this line and use predicted_structure below

policytool/refparse/algo_evaluation/evaluate_parse.py
```diff
@@ -1,34 +1,87 @@
...
-    'Title': parse_test_data['Title'],
-    'PubYear': parse_test_data['PubYear']
-    }
-
-    test_info = parse_test_data
-    test_scores = evaluate_metric(merged_components, parse_test_data['Predicted merged components'])
+    component_names = ['Authors', 'Journal', 'Volume', 'Issue', 'Pagination', 'Title', 'PubYear']
+    actual_components_list = []
+    for _,reference_test_data in parse_test_data.iterrows():
+        a = {}
```
nsorros: a is not a good variable name

policytool/refparse/algo_evaluation/evaluate_parse.py
```diff
@@ -1,34 +1,87 @@
...
+        for component_name in component_names:
+            a.update({component_name : reference_test_data[component_name]})
+        actual_components_list.append(a)
```
nsorros: do you need the actual components_list? why not just pass parse_test_data[category_names] below?

policytool/refparse/algo_evaluation/evaluate_parse.py
```diff
@@ -1,34 +1,87 @@
...
+
+    return lev_dist
+
+def evaluate_metric(actual_components_list, predicted_components_list, levenshtein_threshold):
+
+    # For the computations below it's useful for any nan's to be blank strings
+    df1 = pd.DataFrame(actual_components_list).replace(np.nan,'').astype(str)
+    df2 = pd.DataFrame(predicted_components_list).replace(np.nan,'').astype(str)
+
+    df1['PubYear'] = [d[0:4] if d!='' else d for d in df1['PubYear']]
```
nsorros: the first three lines need to happen before this is called the function should receive the data structured ready

policytool/refparse/algo_evaluation/evaluate_parse.py
```diff
@@ -1,34 +1,87 @@
...
 
-def evaluate_metric(actual, predicted):
+    equal = df1 == df2
 
-    test_scores = {'Score' : 50, 'More info' : "More information about this test"}
+    proportion_equal_cat = equal.sum()/len(equal)
```
nsorros: the naming needs to change instead of proportion let's use accuracy. you need to feel comfortable using machine learning terminology and it is also going to be easier to communicate your metrics if they are standard metrics

policytool/refparse/algo_evaluation/evaluate_parse.py
```diff
@@ -1,34 +1,87 @@
...
+    proportion_equal = (equal.sum().sum())/(equal.shape[0]*equal.shape[1])
```
nsorros: `equal.values.sum() / len(equal.values)` is more readable. also leave spaces in between the `/`

policytool/refparse/algo_evaluation/evaluate_parse.py
```diff
@@ -1,34 +1,87 @@
...
+
+    lev_dist = get_levenshtein_df(df1, df2)
+
+    quite_equal = lev_dist<levenshtein_threshold
+    proportion_quite_equal_cat = quite_equal.sum()/len(quite_equal)
+    proportion_quite_equal = (quite_equal.sum().sum())/(quite_equal.shape[0]*quite_equal.shape[1])
```
nsorros: same changes here. proportion to accuracy, space in between division and different way to divide as above

policytool/refparse/algo_evaluation/evaluate_parse.py
```diff
@@ -1,34 +1,87 @@
...
+
+    test_scores = {
+            'Score' : proportion_equal,
+            'Proportion of components predicted correctly (macro)'\
```
nsorros: You have mixed up what macro and micro means.

Let's define some terminology to use for this code and maybe add it to the top of the document.
Macro: calculate metric per category and then average
Micro: calculate metric for all values irrespective of category

Accuracy: what you call proportion, essentially how many times two arrays agree with each other

Another useful terminology would be the following
Strict: the values we compare to be exactly the same
Lenient: the values we compare to be kind of the same

Use this terminology for defining the metrics and variables. I am expecting to see the following metrics:
* Strict accuracy (macro)
* Strict accuracy (micro)
* Strict accuracy per category
* Lenient accuracy (macro)
* Lenient accuracy (micro)
* Lenient accuracy per category

policytool/refparse/algo_evaluation/evaluate_parse.py
```diff
@@ -1,34 +1,87 @@
...
+    """
+    Input: two equally sized dataframes
+    output: a dataframe of the levenshtein distances between each element pair
+    """
+
+    calc_lev_dist_column = lambda s1, s2:\
+                            editdistance.eval(s1, s2)/max(len(s1), len(s2))\
+                            if (s1!='' or s2 !='')\
+                            else 0
+    lev_dist = pd.DataFrame(
```
nsorros: this function is very complicated. needs rewriting.

one way would be to stack all columns in two big arrays of strings to compare and then calculate distance for each pair

policytool/refparse/algo_evaluation/evaluate_parse.py
```diff
@@ -1,34 +1,87 @@
 
+import pandas as pd
+import numpy as np
+import editdistance
+
 from utils.parse import structure_reference
+from sklearn.metrics.pairwise import cosine_similarity
```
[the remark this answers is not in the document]
lizgzil: @hblanks when you say '1st import' do you mean you just need the full path for the first local file you import?, so would you do e.g.
```
from policytool.refparse.utils.parse import structure_reference
from utils.parse import next_function
```

unpinned_requirements.txt
```diff
@@ -3,6 +3,7 @@ beautifulsoup4
 boto3
 cryptography
 celery
+editdistance
```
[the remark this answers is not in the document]
lizgzil: @SamDepardieu @hblanks any ideas for this?
I tried doing what hunter said and got this error (same error if I deactivate the virtualenv):
```
(virtualenv) 01763-IA-GallaghE:policytool gallaghe$ pwd
/Users/gallaghe/Code/policytool
(virtualenv) 01763-IA-GallaghE:policytool gallaghe$ make update-requirements-txt
if ! [ -d /tmp/update-requirements-txt ]; then \
		mkdir -p /tmp/update-requirements-txt && \
		virtualenv --python python3.6 /tmp/update-requirements-txt && \
		/tmp/update-requirements-txt/bin/pip3 install -r unpinned_requirements.txt ; \
	fi
echo "# Created by 'make update-requirements-txt'. DO NOT EDIT!" > requirements.txt
/tmp/update-requirements-txt/bin/pip freeze | grep -v pkg-resources==0.0.0 | \
		sed 's/airflow/airflow[celery]/' >> requirements.txt
/bin/sh: /tmp/update-requirements-txt/bin/pip: No such file or directory
```

lizgzil:
@nsorros I've addressed your comments.

Only thing to do is to update `requirements.txt`

policytool/refparse/algo_evaluation/evaluate_parse.py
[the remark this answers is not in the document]
nsorros: i think he means generally

policytool/refparse/algo_evaluation/evaluate_parse.py
```diff
@@ -1,34 +1,84 @@
...
+        'Score' : accuracy,
+        'Strict accuracy (micro)': accuracy,
+        'Strict accuracy (per category)': accuracy_cat,
+        'Lenient accuracy (micro)\
+        (normalised Levenshtein < {})'.format(levenshtein_threshold) :\
+            accuracy_quite_equal,
+        'Lenient accuracy (per category)\
+        (normalised Levenshtein < {}) ()'.format(levenshtein_threshold) :\
+            accuracy_quite_equal_cat,
+        'Mean normalised Levenshtein distance (all categories)' :\
```
nsorros: why normalised?

Makefile
```diff
@@ -97,11 +97,11 @@ virtualenv: $(VIRTUALENV)/.installed
...
 update-requirements-txt:
-	if ! [ -d $(VIRTUALENV) ]; then \
-		mkdir -p $(VIRTUALENV) && \
-		virtualenv --python python3.6 $(VIRTUALENV) && \
-		$(VIRTUALENV)/bin/pip3 install -r unpinned_requirements.txt ; \
+	if [ -d $(VIRTUALENV) ]; then \
+		rm -rf $(VIRTUALENV); \
 	fi
+	virtualenv --python python3 $(VIRTUALENV)
+	$(VIRTUALENV)/bin/pip3 install -r unpinned_requirements.txt
```
nsorros: i assume these changes happened with hunter?

policytool/refparse/algo_evaluation/evaluate_parse.py
```diff
@@ -1,34 +1,84 @@
...
+from policytool.refparse.utils.parse import structure_reference
 
-    test_scores = {'Score' : 50, 'More info' : "More information about this test"}
+def get_levenshtein_df(df1, df2):
+    """
+    Input: two equally sized dataframes
+    output: a dataframe of the levenshtein distances between each element pair
+    """
+
+    calc_lev_dist_column = lambda s1, s2:\
```
nsorros: that is more readable i think
```
def calc_lev_distance(s1, s2):
    if s1=='' or s2 =='':
        return 0
    return editdistance.eval(s1, s2) / max(len(s1), len(s2))
```

nsorros: notice you can have a function definition inside a function definition, you do not need to use lambda.

policytool/refparse/algo_evaluation/evaluate_parse.py
nsorros: i guess because you are dividing by the length in your definition

policytool/refparse/algo_evaluation/evaluate_parse.py
```diff
@@ -1,34 +1,84 @@
...
+        'Strict accuracy (micro)': accuracy,
+        'Strict accuracy (per category)': accuracy_cat,
+        'Lenient accuracy (micro)\
+        (normalised Levenshtein < {})'.format(levenshtein_threshold) :\
+            accuracy_quite_equal,
+        'Lenient accuracy (per category)\
+        (normalised Levenshtein < {}) ()'.format(levenshtein_threshold) :\
+            accuracy_quite_equal_cat,
+        'Mean normalised Levenshtein distance (all categories)' :\
+            lev_dist.mean().mean(),
```
nsorros: Notice that this is a macro average

policytool/refparse/algo_evaluation/evaluate_parse.py
lizgzil: yep

Makefile
lizgzil: they did

lizgzil:
@hblanks please could you confirm you are happy with my changes to the requirements files and the makefile changes? The changes are just what we did together.

unpinned_requirements.txt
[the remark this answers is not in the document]
hblanks: Indeed - we fixed this in person yesterday.

PR #150, opened 2019-05-15
Add evaluate match refs metric

lizgzil:
# Description
Similarly to the other PRs, adding the metric to `evaluate_match_references.py`. 

Since we fixed the issue of the fuzzy matcher returning multiple matches, some of `evaluate_match_references` was unnecessarily complicated for the new output of match_vectorised, so I rewrote some of it.

I was also not taking away the negative examples from the match data, so i fixed this.

Also added some info about how we got the evaluation data to the document.

The output of this test is v good, as we might expect:

```
Score
0.998

Micro average F1-score
0.998

Classification report
              precision    recall  f1-score   support

           0       1.00      1.00      1.00       202
           1       1.00      0.99      1.00       199

   micro avg       1.00      1.00      1.00       401
   macro avg       1.00      1.00      1.00       401
weighted avg       1.00      1.00      1.00       401

```
## Type of change

- [ ] :sparkles: New feature

# How Has This Been Tested?

No tests.

# Checklist:

### DIFF
--- a/policytool/refparse/algo_evaluation/evaluate_match_references.py
+++ b/policytool/refparse/algo_evaluation/evaluate_match_references.py
@@ -6,43 +6,48 @@ from utils import FuzzyMatcher
 
 def evaluate_metric(actual, predicted):
 
-    similarity = round(f1_score(actual, predicted, average='micro'), 3)
+    actual_binary = actual.astype(int)
+    # predicted is True, False and "No match found",
+    # convert to 1 (True) or 0 (False or "No match found")
+    predicted_binary = [1 if pred_type==True else 0 for pred_type in predicted]
+
+    similarity = round(f1_score(actual_binary, predicted_binary, average='micro'), 3)
 
     test_scores = {
         'Score' : similarity,
         'Micro average F1-score' : similarity,
-        'Classification report' : classification_report(actual, predicted) 
+        'Classification report' : classification_report(actual_binary, predicted_binary),
+        'Frequency table of match types' : pd.crosstab(index=actual, columns=predicted) 
         }
 
     return test_scores
 
-def evaluate_match_references(match_publications, test_publications, match_threshold):
+def evaluate_match_references(publications, evaluation_references, match_threshold):
+    """
+    Try to match the evaluation_references with references in 'publications'
+    """
 
     # Get the data into the same format as would be
-    test_publications = test_publications.rename(columns={'Match title': 'Title', 'Match pub id': 'Reference id'})
-    test_publications['Journal'] = None
-    test_publications['PubYear'] = None
-    test_publications['Authors'] = None
-    test_publications['Issue'] = None
-    test_publications['Volume'] = None
-    test_publications['Document id'] = range(0, len(test_publications))
-    test_publications['Document uri'] = None
+    evaluation_references = evaluation_references.rename(columns={'Match title': 'Title', 'Match pub id': 'Reference id'})
+    evaluation_references['Document id'] = range(0, len(evaluation_references))
 
     # Take out the negative reference id's (so you shouldn't find them)
-    neg_test_pub_ids = test_publications.loc[test_publications['Type']==0]['Reference id']
-    match_publications_no_negs = match_publications.loc[~match_publications['uber_id'].isin(neg_test_pub_ids)]
+    neg_test_pub_ids = evaluation_references.loc[evaluation_references['Do we expect a match?']==0]['Reference id']
+    publications_no_negs = publications.loc[~publications['uber_id'].isin(neg_test_pub_ids)]
 
-    fuzzy_matcher = FuzzyMatcher(match_publications_no_negs, match_threshold)
+    fuzzy_matcher = FuzzyMatcher(publications_no_negs, match_threshold)
 
-    match_data = fuzzy_matcher.match_vectorised(test_publications)
+    matched_publications = fuzzy_matcher.match_vectorised(evaluation_references)
 
-    # Does each publication have the correct match (True) or not (False)
-    # (There might not be matches found for all the test_publications)
-    merged_test_pred = pd.merge(test_publications, match_data, how="left", left_on='Reference id', right_on='Reference id')
-    match_correct = merged_test_pred['Reference id']==merged_test_pred['uber_id']
+    evaluation_matches = pd.merge(evaluation_references, matched_publications, how="left", left_on='Reference id', right_on='Reference id')
 
-    merged_test_pred['Matched correctly?'] = match_correct
+    # Does each publication have the correct match (true), an incorrect match (false) or no match
+    evaluation_matches['Matched correctly?'] = [
+        matches['Reference id']==matches['uber_id'] if not pd.isnull(matches['uber_id'])\
+        else "No match found"\
+        for _, matches in evaluation_matches.iterrows()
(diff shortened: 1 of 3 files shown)


nsorros:
When I run `evaluate_algo` I get `ValueError: invalid literal for int() with base 10: 'Positive'`

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -24,23 +28,21 @@ def evaluate_match_references(match_publications, test_publications, match_thres
...
-            match_correct.append(1)
-    test_publications['Matched correctly?'] = match_correct
-
-    test_scores = evaluate_metric(test_publications['Type'], test_publications['Matched correctly?'])
+    # Does each publication have the correct match (True) or not (False)
+    # (There might not be matches found for all the test_publications)
+    merged_test_pred = pd.merge(test_publications, match_data, how="left", left_on='Reference id', right_on='Reference id')
+    match_correct = merged_test_pred['Reference id']==merged_test_pred['uber_id']
+
+    merged_test_pred['Matched correctly?'] = match_correct
```
nsorros: redundant line you cna assign above

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -24,23 +28,21 @@ def evaluate_match_references(match_publications, test_publications, match_thres
     test_publications['Document id'] = range(0, len(test_publications))
     test_publications['Document uri'] = None
```
nsorros: why do you need to init all those fields?

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -24,23 +28,21 @@ def evaluate_match_references(match_publications, test_publications, match_thres
...
-            match_correct.append(0)
-        else:
-            match_correct.append(1)
-    test_publications['Matched correctly?'] = match_correct
-
-    test_scores = evaluate_metric(test_publications['Type'], test_publications['Matched correctly?'])
+    # Does each publication have the correct match (True) or not (False)
+    # (There might not be matches found for all the test_publications)
+    merged_test_pred = pd.merge(test_publications, match_data, how="left", left_on='Reference id', right_on='Reference id')
+    match_correct = merged_test_pred['Reference id']==merged_test_pred['uber_id']
```
nsorros: where is there is no match, does this compare nan with nan?

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -24,23 +28,21 @@ def evaluate_match_references(match_publications, test_publications, match_thres
...
-        if (not uberids) or (test_publication['Reference id'] not in uberids):
-            #  If there were no matches found for this reference or the id of the match was different
-            match_correct.append(0)
-        else:
-            match_correct.append(1)
-    test_publications['Matched correctly?'] = match_correct
-
-    test_scores = evaluate_metric(test_publications['Type'], test_publications['Matched correctly?'])
+    # Does each publication have the correct match (True) or not (False)
+    # (There might not be matches found for all the test_publications)
```
nsorros: Not **there might** they **should not** unless i am not understanding something here. we merge back because we know and expect the negatives not to match

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -24,23 +28,21 @@ def evaluate_match_references(match_publications, test_publications, match_thres
...
+    merged_test_pred = pd.merge(test_publications, match_data, how="left", left_on='Reference id', right_on='Reference id')
+    match_correct = merged_test_pred['Reference id']==merged_test_pred['uber_id']
+
+    merged_test_pred['Matched correctly?'] = match_correct
+
+    test_scores = evaluate_metric(merged_test_pred['Type'].astype(int), merged_test_pred['Matched correctly?'].astype(int))
```
nsorros: `merged_test_pred['Type']` Type is not very informative, can we find a different name? Is this column suppose to represent whether we should find a match or not?

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -24,23 +28,21 @@ def evaluate_match_references(match_publications, test_publications, match_thres
     test_publications['Document id'] = range(0, len(test_publications))
     test_publications['Document uri'] = None
 
-    fuzzy_matcher = FuzzyMatcher(match_publications, match_threshold)
+    # Take out the negative reference id's (so you shouldn't find them)
+    neg_test_pub_ids = test_publications.loc[test_publications['Type']==0]['Reference id']
+    match_publications_no_negs = match_publications.loc[~match_publications['uber_id'].isin(neg_test_pub_ids)]
+
+    fuzzy_matcher = FuzzyMatcher(match_publications_no_negs, match_threshold)
```
nsorros: the variable names are bit confusing to follow. maybe it would be better to name the publications we are matching against publications and the ones we want to match references. otherwise we name the ones publications, same as the previous and the other publications_to_match or query_publications

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -24,23 +28,21 @@ def evaluate_match_references(match_publications, test_publications, match_thres
...
 
     match_data = fuzzy_matcher.match_vectorised(test_publications)
```
nsorros: if we move on with renaming then this could be named matched_pulbications which is more appropriate

lizgzil:
> When I run `evaluate_algo` I get `ValueError: invalid literal for int() with base 10: 'Positive'`

download "positive_and_negative_match_test_data.csv" again from S3, I have changed 'positive' and 'negative' to 1 and 0, but forgot to update S3

nsorros:
> > When I run `evaluate_algo` I get `ValueError: invalid literal for int() with base 10: 'Positive'`
> 
> download "positive_and_negative_match_test_data.csv" again from S3, I have changed 'positive' and 'negative' to 1 and 0, but forgot to update S3

It worked 👍

policytool/refparse/algo_evaluation/evaluate_match_references.py
lizgzil: 👍  I thought it would complain, but it actually didn't!

policytool/refparse/algo_evaluation/evaluate_match_references.py
lizgzil: `merged_test_pred['Reference id']` will always have an id (it's the input data), but when there is no match it is true that `merged_test_pred['uber_id']` will be nan. So we never compare nan with nan, but we do have e.g. `'pub.1019573229' == nan`.

So `merged_test_pred['Matched correctly?']` being true means the reference was correctly matched, but being false means either it was matched to a different reference than itself, or no match was found.

lizgzil: I'm thinking it might be useful to return correct match, incorrect match and no match.

policytool/refparse/algo_evaluation/evaluate_match_references.py
lizgzil: changed to 'Do we expect a match?'

PR #151, opened 2019-05-16
Update merge results to return predicted references

nsorros:
# Description

This PR adds functionality for merge_results to return the predicted references as well as the matches which is the other half of the output of the policy tool. This became necessary after realizing that the database is not maintained and it is a better idea to rely on runs of the policy tool for getting the latest numbers than querying the database.

## Type of change

- [x] :sparkles: New feature

# How Has This Been Tested?

`make docker-test`

# Checklist:

### DIFF
--- a/policytool/refparse/merge_results.py
+++ b/policytool/refparse/merge_results.py
@@ -64,7 +64,6 @@ def concat_predicted_csvs(predicted_csv_names):
         all_url - a dataframe containing document id and document url
     """
     all_predicted_references = []
-    all_url = []
     for predicted_csv_name in predicted_csv_names:
         # Some (4) of the files don't read in without errors
         try:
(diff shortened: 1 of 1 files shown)


policytool/refparse/merge_results.py
```diff
@@ -40,7 +40,11 @@ def concat_match_csvs(match_csv_names):
     """
     all_matches = []
     for match_csv_name in match_csv_names:
-        match_data = pd.read_csv(match_csv_name)
+        try:
+            match_data = pd.read_csv(match_csv_name)
+        except:
+            print(match_csv_name)
```
nsorros: oups

nsorros:
This code does not run after the remove i think due to `from .refparse import get_file`. It complains that ref parse does not have `__main__` and is not package. Removing the dot makes the refparse complain that has imports like that. I would like to discuss this with @hblanks and @SamDepardieu to understand what is going on.

hblanks:
> This code does not run after the remove i think due to from .refparse import get_file. It complains that ref parse does not have __main__ and is not package. Removing the dot makes the refparse complain that has imports like that. I would like to discuss this with @hblanks and @SamDepardieu to understand what is going on.

@nsorros  - you may want to try:

```
python -m policytool.refparse.merge_results
```

Instead of what I presume you're running:

```
python policytool/refparse/merge_results.py
```

In short, Python only understands "packages" and thus relative imports when they are invoked as modules, not just plain files somewhere in the filesystem. So the `-m ${MODULE_NAME}` invocation style may solve your problem for you. If not, though, LMK!

nsorros:
I am closing this PR for now as I think the other PR makes merge_results.py obsolete. Happy to reopen it if we think otherwise when I open it for review.

PR #153, opened 2019-05-17
Add evaluate scraper metric

lizgzil:
Note you will need my testing data for this (temporary until I actually get the real policy pdf data). Download the 'pdf_sections' and 'pdfs' folders in https://s3.console.aws.amazon.com/s3/buckets/datalabs-data/policy_tool_tests/?region=us-east-2&tab=overview and put them in the 'data_evaluate' folder.

# TO DO 

- Get the actual evaluation data! (currently test pdfs) (can still review this PR without this though, but would be nice to check it doesn't fail or anything once we have the proper data)

# Description

This involved significant work at working out how the test would work, so I have had to redesign what the evaluation data looks like, and import the new type and process accordingly (in `evaluate_algo.py`).

I have rewritten most of the layout I had designed in `evaluate_find_section.py`, and imported different files (editing `evaluate_settings.py`). I've also added a bit of text to the README about importing poppler to make this evaluation work.

I also found an error in `../pdf_parser/main.py` - the file path no longer existed, so I've corrected this file path.

## Type of change

- [x] :sparkles: New feature
- [ ] :fire: Breaking change

# How Has This Been Tested?

No tests.

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,11 @@ This will test that your last changes didn't affect how the program works.
 
 ## Evaluating each component of the algorithm
 
-We have devised some evaluation data in order to evaluate 5 steps of the model. The results can be found by first downloading the evaluation data from [here](https://s3-eu-west-1.amazonaws.com/datalabs-data/policy_tool_tests), and then running
+We have devised some evaluation data in order to evaluate 5 steps of the model. The results can be found by first installing poppler
+```
+brew install poppler
+```
+and then downloading the evaluation data from [here](https://s3-eu-west-1.amazonaws.com/datalabs-data/policy_tool_tests), and then running
 ```
 python evaluate_algo.py --verbose True
 ```
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ This will test that your last changes didn't affect how the program works.
 
 ## Evaluating each component of the algorithm
 
-We have devised some evaluation data in order to evaluate 5 steps of the model. The results can be found by first installing poppler
+We have devised some evaluation data in order to evaluate 5 steps of the model. The results can be calculated by first installing poppler
 ```
 brew install poppler
 ```
--- a/policytool/refparse/algo_evaluation/evaluate_find_section.py
+++ b/policytool/refparse/algo_evaluation/evaluate_find_section.py
@@ -34,20 +34,23 @@ def evaluate_metric_scraped(actual, predicted, sections):
 
     test_scores = {
         'Score' : similarity,
-        'Micro average F1-score' : similarity,
+        'F1-score' : similarity,
         'Classification report' : classification_report(actual, predicted),
         'Confusion matrix' : pretty_confusion_matrix(
                 actual, predicted, [True, False]
             )
         }
 
+    sections_texts = pd.DataFrame(
+        {'Section': sections, 'Actual': actual, 'Predicted': predicted}
+        )
+
     for section_name in set(sections):
-        actual_section = [
-            a for (s,a) in zip(sections, actual) if s == section_name
-        ]
-        predicted_section = [
-            p for (s,p) in zip(sections, predicted) if s == section_name
-        ]
+        section_text = sections_texts[sections_texts['Section']==section_name]
+
+        actual_section = section_text['Actual']
+        predicted_section = section_text['Predicted']
+
         test_scores["Classification report for the {} section".format(
             section_name
             )] = classification_report(actual_section, predicted_section)
--- a/policytool/refparse/algo_evaluation/evaluate_find_section.py
+++ b/policytool/refparse/algo_evaluation/evaluate_find_section.py
@@ -1,48 +1,138 @@
+import sys
+sys.path.append("..")
+from pdf_parser.pdf_parse import parse_pdf_document, grab_section
 
-def evaluate_metric_scraped(actual, predicted):
+import os.path
+import pandas as pd
+import numpy as np
+
+from sklearn.metrics import classification_report, f1_score
+import editdistance
+
+def evaluate_metric_scraped(actual, predicted, section_names):
     """
-    False positive rate
+    Input:
+        actual : a boolean list of whether section text was in the pdf
+        predicted : a boolean list of whether section text was scraped
+        section_names : a list of the section names
     """
-    # Place holder
-    test_scores = {'Score' : 50, 'More info' : "More information about this test"}
+
+    similarity = round(f1_score(actual, predicted, average='micro'), 3)
+
+    test_scores = {
+        'Score' : similarity,
+        'Micro average F1-score' : similarity,
+        'Classification report' : classification_report(actual, predicted) 
+        }
+
+    for section_name in set(section_names):
+        has_section = [e for i,e in enumerate(actual) if section_names[i] == section_name]
+        has_scraped_section = [e for i,e in enumerate(predicted) if section_names[i] == section_name]
+        test_scores["Classification report for the {} section".format(section_name)] = classification_report(has_section, has_scraped_section)
 
     return test_scores
 
 
-def evaluate_metric_quality(actual, predicted):
+def evaluate_metric_quality(actual, predicted, sections, levenshtein_threshold):
     """
-    Text similarity
+    Text similarity for when there is a section (actual!='')
     """
-    test_scores = {'Score' : 50, 'More info' : "More information about this test"}
 
+    section_texts_all = [{'Actual section text' : a, 'Predicted section text' : p, 'Section' : s} for a,p,s in zip(actual, predicted, sections)]
+
+    section_texts_all = list(filter(lambda x: x['Actual section text'] != '', section_texts_all))
+
+    lev_distance = []
+    for section_texts in section_texts_all:
+        actual_text = section_texts['Actual section text']
+        predicted_text = section_texts['Predicted section text']
+        lev_distance.append(editdistance.eval(actual_text, predicted_text) / max(len(actual_text), len(predicted_text)))
+
+    # Sections for those that actually exist
+    sections_exist = [s['Section'] for s in section_texts_all]
+
+    # Which sections (of the texts that actual exist) were found exactly?
+    equal = [l==0 for l in lev_distance]
+    
+    # Which sections (of the texts that actual exist) were found roughly the same?
+    quite_equal = [l<levenshtein_threshold  for l in lev_distance]
+
+    test_scores = {
+        'Score' : sum(equal)/len(equal),
+        'Mean normalised Levenshtein distance' : np.mean(lev_distance),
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -105,7 +105,11 @@ This will test that your last changes didn't affect how the program works.
 
 ## Evaluating each component of the algorithm
 
-We have devised some evaluation data in order to evaluate 5 steps of the model. The results can be found by first downloading the evaluation data from [here](https://s3-eu-west-1.amazonaws.com/datalabs-data/policy_tool_tests), and then running
+We have devised some evaluation data in order to evaluate 5 steps of the model. The results can be found by first installing poppler
```
nsorros: The results will be calculated not found, right?

policytool/refparse/evaluate_algo.py
```diff
@@ -106,31 +176,48 @@ def create_argparser():
     )
 
     # Load the latest parser model
-    model = fm.get_file('reference_parser_pipeline.pkl','./reference_parser_models/','pickle')
+    model = fm.get_file(
+        'reference_parser_pipeline.pkl',
+        './reference_parser_models/',
```
nsorros: shouldn't those things be in settings?

policytool/refparse/evaluate_algo.py
```diff
@@ -47,49 +88,78 @@ def create_argparser():
...
         settings.FOLDER_PREFIX,
-        'csv'
+        settings.SCRAPE_DATA_PDF_FOLDER_NAME
+    )
+
+    pdf_names = listdir(scrape_pdf_location)
+    pdf_names = [
+        os.path.splitext(pdf_name)[0] for pdf_name in pdf_names
+    ]
+    pdf_names.remove('.DS_Store')
```
nsorros: this should not be in the code

policytool/refparse/evaluate_algo.py
```diff
@@ -4,32 +4,73 @@
...
+    For a pdf name in the evaluation data, 
+    get all the text from any relevant txt files in the locations given
+    
+    Input:
+    pdf_name : A pdf name in the evaluation data 
+                e.g pdf_name1
+    sections_folder_names : 
+                A list of the folder names for each
+                section we have evaluation data for
+                e.g. ['reference', 'bibliograph']
```
nsorros: the example needs to be an example of sections_folder_names not sections locations

policytool/refparse/evaluate_algo.py
```diff
@@ -4,32 +4,73 @@
...
+        pdf_name, sections_folder_names, sections_location
+        ):
+    """
+    For a pdf name in the evaluation data, 
+    get all the text from any relevant txt files in the locations given
+    
+    Input:
+    pdf_name : A pdf name in the evaluation data 
+                e.g pdf_name1
+    sections_folder_names :
```
nsorros: section folder names is slightly confusing. section names is a better name, since this is what it is regardless of how we have implemented the saving which in this case is one folder per section name, in the future it can all be in one json

policytool/refparse/evaluate_algo.py
```diff
@@ -47,49 +88,78 @@ def create_argparser():
...
+    sections_folder_names = [
+        name for name in os.listdir(sections_location) \
+        if os.path.isdir('{}/{}'.format(sections_location, name))
+    ]
+
+    evaluate_find_section_data = {
+        pdf_name : get_pdf_sections(
+            pdf_name, sections_folder_names, sections_location
+        ) for pdf_name in pdf_names
+    }
```
nsorros: I personally find something like this more readable.
```
from collections import defaultdict

def get_text(filepath):
    try:
        references_section = open(filepath).read()
    except:
        references_section = ""

    return references_section

def yield_section_data(scrape_pdf_location, sections_location):
    sections_names = [
        section_name
        for section_name in os.listdir(sections_location)
    ]
    for filename in listdir(scrape_pdf_location):
        pdf_hash, _ = os.path.splitext(filename)
        for section_name in sections_names:
            section_path = os.path.join(sections_location, section_name)
            section_text = get_text(section_path)
            yield pdf_hash, section_name, section_text

evaluate_find_section_data = defaultdict(defaultdict(str))
for pdf_hash, section_name, section_text in yield_section_data(scrape_pdf_location, sections_location):
    evaluate_find_section_data[pdf_hash][section_name] = section_text
```

policytool/refparse/evaluate_algo.py
```diff
@@ -47,49 +88,78 @@ def create_argparser():
...
+
+    sections_location = '{}/{}'.format(
+        settings.FOLDER_PREFIX,
+        settings.SCRAPE_DATA_REF_PDF_FOLDER_NAME
     )
 
-    # Load data to test split sections for test 3:
+    sections_folder_names = [
+        name for name in os.listdir(sections_location) \
+        if os.path.isdir('{}/{}'.format(sections_location, name))
```
nsorros: ideally this would not be here as you control whether the folder contains dirs and if it does not that is an error on the data. i would suggest removing that and constructing the folder structure accordingly, i.e. without files and folders, only folders

policytool/refparse/evaluate_algo.py
```diff
@@ -47,49 +88,78 @@ def create_argparser():
...
+        settings.SCRAPE_DATA_PDF_FOLDER_NAME
+    )
+
+    pdf_names = listdir(scrape_pdf_location)
+    pdf_names = [
+        os.path.splitext(pdf_name)[0] for pdf_name in pdf_names
+    ]
+    pdf_names.remove('.DS_Store')
+
+    sections_location = '{}/{}'.format(
```
nsorros: it is better to use `os.path.join` because it is cross platform, i.e. it works in windows and unix whereas this way would not work on windows

policytool/refparse/evaluate_algo.py
```diff
@@ -47,49 +88,78 @@ def create_argparser():
...
+        settings.FOLDER_PREFIX,
+        settings.SCRAPE_DATA_REF_PDF_FOLDER_NAME
     )
 
-    # Load data to test split sections for test 3:
+    sections_folder_names = [
+        name for name in os.listdir(sections_location) \
```
nsorros: i think you import listdir directly so you do not need to do `os.listdir` just `listdir`

policytool/refparse/algo_evaluation/evaluate_find_section.py
```diff
@@ -1,48 +1,176 @@
...
+    return scrape_data
+
+
+def evaluate_find_section(
+        evaluate_find_section_data, scrape_pdf_location, levenshtein_threshold
+        ):
+
+    # Get the predicted text for each of the pdf sections for each pdf
+    section_names = evaluate_find_section_data[
+        next(iter(evaluate_find_section_data))
```
nsorros: what is going on here?

policytool/refparse/algo_evaluation/evaluate_find_section.py
```diff
@@ -1,48 +1,176 @@
...
+                actual, predicted, [True, False]
+            )
+        }
+
+    for section_name in set(sections):
+        actual_section = [
+            a for (s,a) in zip(sections, actual) if s == section_name
+        ]
+        predicted_section = [
+            p for (s,p) in zip(sections, predicted) if s == section_name
```
nsorros: it is an overkill to zip just so that you can filter. plus you are duplicating code. 
```
ap_zip = zip(actual, predicted)
[ap_zip[i] for i, s, in enumerate(sections) if s == section_name]
actual_section, predicted_section(*ap_zip)
```

Note that this is a good excuse to use a dataframe
```
df = pd.DataFrame({'s': sections, 'a': actuals, 'p': predicted})
sdf = df[df['s']==s]
actual_section = sdf['a']
predicted_section = sdf['p']
```

policytool/refparse/algo_evaluation/evaluate_find_section.py
```diff
@@ -1,48 +1,176 @@
...
-        
-        if data['References section']:
-            references_text.append(data['References section']) #### NO IDEA HERE - GET THE REFERENCES SECTION TEXT SCRAPED
-        else:
-            references_text.append(None) # It might be that nothing is scraped
+    similarity = round(f1_score(actual, predicted, average='micro'), 3)
+
+    test_scores = {
+        'Score' : similarity,
+        'Micro average F1-score' : similarity,
```
nsorros: what does micro mean in this case? we have two classes here right? it is not like we have more than one classes and we could macro average them somehow.

policytool/refparse/algo_evaluation/evaluate_find_section.py
```diff
@@ -1,48 +1,176 @@
...
 
-    scrape_test_data['References section text scraped'] = references_text
-
-    return scrape_test_data
+    return test_scores
 
-def evaluate_find_section(scrape_test_data):
 
-    test1_2_info = scrape_pdfs(scrape_test_data)
+def evaluate_metric_quality(scrape_data, levenshtein_threshold):
```
nsorros: this function is very similar to the one we use for evaluate_parse, it would be great to refactor the common parts to a common function that both use

policytool/refparse/algo_evaluation/evaluate_find_section.py
```diff
@@ -1,48 +1,176 @@
...
+
+    test_scores = {
+        'Score' : np.mean(equal),
+        'Mean normalised Levenshtein distance' : np.mean(lev_distances),
+        'Strict accuracy (micro)' : np.mean(equal),
+        'Lenient accuracy (micro)' : np.mean(quite_equal)}
+
+    for section_name in set(sections):
+        # Get the Levenshtein distances for this sections actual-predicted pairs
+        lev_distances_section = [
```
nsorros: you are duplicating a lot of code here with code up. you should not do that. refactor the common functionality to common functions. the best refactor would be to create common functions with ref_parse actually.

policytool/refparse/algo_evaluation/evaluate_find_section.py
lizgzil: getting the names of the keys from the first item in the dictionary (it's a nested dictionary). basically pulling out ['reference', 'bibliograph'] from:

{pdf1: {'reference': '', 'bibliograph' : ''}, pdf2 : ...}

policytool/refparse/algo_evaluation/evaluate_find_section.py
lizgzil: it was just because i'd used the arg average='micro' but yeh it makes no difference and would make the results doc confusing

policytool/refparse/algo_evaluation/evaluate_find_section.py
lizgzil: i will merge this and then work on a PR where I combine everything and set up common functions where there is overlap

PR #156, opened 2019-05-22
Match refs task airflow

nsorros:
This is a big PR. Apologies for this in advance, it was not obvious that it would be that big nor that it will take so much time to complete.

This contains the airflow tasks for matching references and publications using Airflow and Elasticsearch. To make it happen I had to
* refactor refparse, mainly to make the hard text matcher, now called exact matcher, to work on one publication at a time. i also made fuzzy matcher to work on reference at a time which is how elasticsearch would work.
* implement the operators in Airflow using Elasticsearch which required finding a good enough query to make the functionality similar.
* fill elasticsearch with sections, fulltexts and references to test the functionality
* populate the test and policytool dag

Things that I have not done
* I have not written tests for the new implementation of matching
* I have not managed to run the policytool dag and lately the test dag is not really working
* I have not implemented the functionality in the exact same way as this would require more time so there are some differences
* I have not thoroughly investigated the threshold, I am hoping to use Liz code to decide the threshold

To make sure results are the same or similar I have written a comparison code that compares the old with the new implementation and that shows that they mostly agree.

**Important note to reviewers**
The best and recommended way to review this PR would be to follow the commits. To a certain extent each commit, could be its own PR. The last two commits may be omitted and we might decide to drop them as they implement functionality to load the local elasticsearch and test the dags. The main changes are in the first three commits.

**Import data to Elasticsearch**
```
python policytool/elastic/import_sections_from_s3.py s3://datalabs-data/scraper-results/msf/20190117.json
python policytool/elastic/import_refs_from_s3.py s3://datalabs-data/reference_parser/data/raw/uber_api_publications.csv
```


policytool/refparse/refparse.py
```diff
@@ -207,66 +184,94 @@ def parse_references(scraper_file, model_file,
...
-                    )
-                )
-            matched_references.to_csv(
-                os.path.join(
-                    output_dir,
-                    f"{doc_id}_{settings.MATCHES_FILENAME}"
+    with open(structured_references_filepath, 'w') as srefs_f:
+        with open(fuzzy_matched_references_filepath, 'w') as fmrefs_f:
+
+            refs = parse_references(
```
lizgzil: I think `refs` might need a rename now, e.g. something to convey it is for all the documents. `documents_structured_references`
`structured_reference` in `structured_references` in `refs`,
`structured_reference` in `structured_references_doc` in `structured_references_docs`.
A bit confusing either way I suppose

policytool/refparse/refparse.py
```diff
@@ -207,66 +184,94 @@ def parse_references(scraper_file, model_file,
...
 
+    scraper_file = get_file(scraper_file, "", get_scraped=True)
+    sectioned_documents = transform_scraper_file(scraper_file)
+    exact_matcher = ExactMatcher(sectioned_documents, settings.MATCH_TITLE_LENGTH_THRESHOLD)
+    with open(exact_matched_reference_filepath, 'w') as emrefs_f: 
+        for publication in publications:
+            exact_matched_references = exact_match_publication(exact_matcher, publication)
+            for exact_matched_reference in exact_matched_references:
+                if exact_matched_reference:
+                    emrefs_f.write(json.dumps(exact_matched_reference)+'\n')
```
lizgzil: What is the benefit to outputting both the fuzzy matched and the exact matched separately? Is it not a waste of space since there will be duplications?

policytool/refparse/refparse.py
```diff
@@ -207,66 +184,94 @@ def parse_references(scraper_file, model_file,
...
+                    if structured_reference:
+                        srefs_f.write(json.dumps(structured_reference)+'\n')
 
+    scraper_file = get_file(scraper_file, "", get_scraped=True)
+    sectioned_documents = transform_scraper_file(scraper_file)
+    exact_matcher = ExactMatcher(sectioned_documents, settings.MATCH_TITLE_LENGTH_THRESHOLD)
+    with open(exact_matched_reference_filepath, 'w') as emrefs_f: 
+        for publication in publications:
+            exact_matched_references = exact_match_publication(exact_matcher, publication)
+            for exact_matched_reference in exact_matched_references:
```
lizgzil: Is there not a way to json.dumps many references at one at a time?

policytool/refparse/utils/fuzzy_match.py
```diff
@@ -0,0 +1,51 @@
...
+        retrieved_publications.sort_values(by='similarity', ascending=False, inplace=True)
+        return retrieved_publications[:nb_results]
+
+    def match(self, reference):
+        if len(reference['Title']) < self.title_length_threshold:
+            return
+
+        retrieved_publications = self.search_publications(reference)
+
+        best_match = retrieved_publications.iloc[0]
```
lizgzil: I think it's important that we pick a random top match (if there are multiple with the same cosine). I guess my old way of doing it was a bit messy, perhaps it'd be best to randomly shuffle the publications before we do any of the matching. 
Could do `self.publications = pd.DataFrame(publications).sample(frac=1)` in line 12? (https://stackoverflow.com/questions/29576430/shuffle-dataframe-rows)

policytool/airflow/tasks/fuzzy_match_refs_operator.py
```diff
@@ -0,0 +1,135 @@
...
+        key.download_fileobj(tf)
+
+        tf.seek(0)
+
+        with gzip.GzipFile(mode='rb', fileobj=tf) as f:
+            for line in f:
+                yield json.loads(line)
+
+class ElasticsearchFuzzyMatcher:
+    def __init__(self, es, score_threshold, should_match_threshold, title_length_threshold=0):
```
lizgzil: Is there a reason to use `score_threshold` instead of `similarity_threshold` like you call it in fuzzy matcher?

policytool/airflow/tasks/fuzzy_match_refs_operator.py
```diff
@@ -0,0 +1,135 @@
...
+            size=1
+        )
+
+        matches_count = res['hits']['total']['value']
+        if matches_count == 0:
+            return
+        
+        best_match = res['hits']['hits'][0]
+        best_score = best_match['_score']
+        if best_score > self.score_threshold:
```
lizgzil: this code is from fuzzymatcher, so shouldnt you import it and use it rather than copying the code here?

policytool/airflow/tasks/exact_match_refs_operator.py
```diff
@@ -0,0 +1,127 @@
...
+            for line in f:
+                yield json.loads(line)
+ 
+class ElasticsearchExactMatcher:
+    def __init__(self, es, title_length_threshold):
+        self.es = es
+        self.title_length_threshold = title_length_threshold
+
+    def match(self, publication):
+        title = publication['title']
```
lizgzil: how come you don't clean the text like you do in `exact_match.py`?

policytool/airflow/dags/policytool_dag.py
```diff
@@ -20,6 +23,8 @@
     'msf',
 ]
 
+MIN_TITLE_LEGTH = 40
```
lizgzil: spelling mistake

policytool/airflow/dags/test_dag.py
```diff
@@ -29,45 +23,21 @@
...
-extracted_refs_path = os.path.join(
-        'datalabs-data',
-        'airflow',
-        'output',
-        'policytool-extract',
-        'test-extract-refs-{organisation}.json.gz'.format(
-            organisation=organisation
-        )
-        )
+fuzzy_match_refs = FuzzyMatchRefsOperator(
```
lizgzil: I guess you didnt want to run all the other operators not relevant to the fuzzymatch operator in the test, but did you mean to leave them deleted? not sure how this dag should be left

policytool/airflow/dags/policytool_dag.py
```diff
@@ -74,23 +80,57 @@
...
+        dag=dag
+    )
     extract_refs.set_upstream(pdfParsing)
 
+    fm_references_path = os.path.join(
+        output_path
+        'policytool-extract',
+        'fuzzy-match-refs-{organisation}.json.gz'.format(organisation=organisation)
+    )
+    fmMatching = FuzztMatchRefsOperator(
```
lizgzil: I thought the input to this operator should be the references extracted from the previous step, so I am confused to not see the `extracted_refs_path` as an input?

lizgzil: spelling mistake (FuzztMatchRefsOperator) might be the reason you cant run it!

policytool/airflow/dags/policytool_dag.py
```diff
@@ -74,23 +80,57 @@
...
+        task_id='fuzzy_match_refs',
+        es_host='http://elasticsearch:9200',
+        structured_references_path=output_path,
+        fuzzy_matched_references_path=fm_references_path,
+        dag=dag
+    )
+    fmMatching.set_upstream(extractRefs)
+    for year in range(2000,2019):
```
lizgzil: how come you do the exact matcher year by year?

policytool/airflow/dags/policytool_dag.py
```diff
@@ -74,23 +80,57 @@
...
+            output_path,
+            'open-research',
+            'dimensions',
+            'publications'
+            'dimensions-publications-{year}.json.gz'.format(year=year)
+        )
+        emMatching = ExactMatchRefsOperator(
+            task_id='hard_match_refs',
+            es_host='http://elasticsearch:9200',
+            publications_path=publications_path,
```
lizgzil: again, shouldn't the input be the output of the `ParsePdfOperator`, i.e. `parsing_path` instead of `publications_path`?

policytool/airflow/dags/policytool_dag.py
lizgzil: i suppose if you used the input to be `parsing_path` youd want to loop through organisation rather than by year

policytool/elastic/import_refs_from_s3.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def clean_es(es):
+    """ Empty the elasticsearch database.
+
+    Args:
+        es: a living connection to elasticsearch
+
+    """
+    print('Cleaning te database..')
```
lizgzil: spelling mistake (?)

policytool/airflow/tasks/fuzzy_match_refs_operator.py
```diff
@@ -0,0 +1,135 @@
...
+        with gzip.GzipFile(mode='rb', fileobj=tf) as f:
+            for line in f:
+                yield json.loads(line)
+
+class ElasticsearchFuzzyMatcher:
+    def __init__(self, es, score_threshold, should_match_threshold, title_length_threshold=0):
+        self.es = es
+        self.score_threshold = score_threshold
+        self.title_length_threshold = title_length_threshold
+        self.should_match_threshold = should_match_threshold
```
lizgzil: do you say anywhere what `should_match_threshold` is? because I'm not sure what it is

lizgzil:
I have left some comments.
General points:
- I wondered why you didn't use some of the functions from the fuzzymatcher and exact matcher code, it seems perhaps a bit unideal to have 2 copies of them. I imagine this is something to do with the format elastic search needs them in?
- I am also a bit confused about the flow of the policy dag, which I thought had to follow the rule that the input of one task was the output of a previous task? Again perhaps this is because I don't know how elasticsearch works enough.
- I couldn't see where you use the functions from `import_refs_from_s3.py` and `import_sections_from_s3.py`? I also have a local file called `import_from_s3.py` which looks very similar to `import_refs_from_s3.py`, do you need to delete it?

policytool/airflow/dags/policytool_dag.py
nsorros: i like how you say spelling mistake but you do not ruin the surprise to where the mistake is so that there is still an element for a surprise :P will correct

policytool/airflow/dags/policytool_dag.py
nsorros: tha parsed text will be indexed to elasticsearch by another task. assuming that it is there though, the exact matcher get a list of publications and runs them against the indexed text

policytool/airflow/dags/test_dag.py
nsorros: i think test_dag is just for testing so i deleted but happy to reintroduce any.

policytool/airflow/tasks/exact_match_refs_operator.py
nsorros: that is a good question. well elasticsearch applies basic cleaning when indexing and before running a query by default. if we wanted to change that we would need to change it in the settings and as part of the querying it would apply it. at the moment i am not applying any cleaning apart from lowercase i think, should we introduce the same logic?

policytool/airflow/tasks/fuzzy_match_refs_operator.py
nsorros: this is meant to communicate that is is a different number. in one case it is the cosine similarity and in the other it is the relevant score that elasticsearch calculated. on the first case it goes from 0 to 1, on the second case it goes from 0 to arbitrary high

policytool/airflow/tasks/fuzzy_match_refs_operator.py
nsorros: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-minimum-should-match.html

should we add an explanation or the link somewhere?

policytool/airflow/tasks/fuzzy_match_refs_operator.py
nsorros: you have a good eye, i have intentionally written them the same way. we could do that yes.

policytool/elastic/import_refs_from_s3.py
nsorros: xixi, indeed

policytool/refparse/refparse.py
nsorros: @hblanks might have a view on this. i am under the impression that one at a time is the right thing to do.

policytool/refparse/refparse.py
nsorros: The deduplication will happen in a different task I guess, that is why. So that task will load both and deduplicate.

policytool/refparse/utils/fuzzy_match.py
nsorros: I would suggest to open an issue to discuss this and include Aoife. After our discussion yesterday, I am thinking that random might be weird in some cases, since different runs of the algorithm will give different results.

policytool/airflow/tasks/exact_match_refs_operator.py
lizgzil: hm well it does feel like it should be lowercase, but if thats included in the basic clean then thats ok?

policytool/airflow/tasks/fuzzy_match_refs_operator.py
lizgzil: oh i see!

policytool/airflow/tasks/fuzzy_match_refs_operator.py
lizgzil: ah I didnt realise it was an elasticsearch parameter. In that case I think it is probably fine to not document it

policytool/refparse/utils/fuzzy_match.py
```diff
@@ -0,0 +1,51 @@
...
+import pandas as pd
+from policytool.refparse.settings import settings
+from sklearn.feature_extraction.text import TfidfVectorizer
+from sklearn.metrics.pairwise import cosine_similarity
+
+
+
+class FuzzyMatcher:
+    def __init__(self, publications, similarity_threshold, title_length_threshold=0):
+        self.logger = settings.logger
```
hblanks: I'd probably just do `logger = logging.getLogger(__name__)` outside of the class and use that instead of `self.logger`.

policytool/refparse/utils/fuzzy_match.py
```diff
@@ -0,0 +1,51 @@
...
+                'Document id': reference['Document id'],
+                'Reference id': reference['Reference id'],
+                'Extracted title': reference['Title'],
+                'Matched title': best_match['title'],
+                'Matched publication id': best_match['uber_id'],
+                'Similarity': best_similarity,
+                'Match algorithm': 'Fuzzy match'
+            }
+
+        return
```
hblanks: Extraneous line

policytool/airflow/tasks/exact_match_refs_operator.py
nsorros: this is the default https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-analyzer.html

policytool/airflow/tasks/fuzzy_match_refs_operator.py
nsorros: more information on that. it is a long read https://www.elastic.co/guide/en/elasticsearch/guide/current/controlling-relevance.html

policytool/airflow/tasks/fuzzy_match_refs_operator.py
nsorros: actually, on a second read, there are slight differences which make it slightly more complicated to refactor. i am referring to the fact that there are no hits, score or source in the local fuzzymatcher.

policytool/airflow/dags/policytool_dag.py
nsorros: You are right, the structured_references_path should be extracted_refs_path, will change.

nsorros:
@lizgzil @hblanks Most code review comments are addressed. This should be ready to merge.

@lizgzil Also the test dag is working if you want to give it a go. I have not tried making the policy dag working as per hunter recommendation.

PR #157, opened 2019-05-23
Extract ref task

lizgzil:
# Description

- Add `extract_refs_operator.py` which inputs parsed pdfs in json form and outputs a json gz of structured references.
- Add the hook folder and the wellcome s3 hook
- Edit `test_dag.py` to include this task
- Edit `policytool_dag.py` to include this task
- Restructure `refparse.py` so that it yields the structured references and then matches them in different functions
- Edit the settings location imports in the utils code (`settings` to `policytool.refparse.settings`) 

## Type of change

- [ ] :sparkles: New feature

# How Has This Been Tested?

unittests succeed but with some warnings.

```
(virtualenv) 01763-IA-GallaghE:policytool gallaghe$ cd refparse/
(virtualenv) 01763-IA-GallaghE:refparse gallaghe$ python -m unittest
/Users/gallaghe/Code/policytool/build/virtualenv/lib/python3.7/site-packages/sklearn/base.py:253: UserWarning: Trying to unpickle estimator CountVectorizer from version 0.19.2 when using version 0.20.3. This might lead to breaking code or invalid results. Use at your own risk.
  UserWarning)
/Users/gallaghe/Code/policytool/build/virtualenv/lib/python3.7/site-packages/sklearn/base.py:253: UserWarning: Trying to unpickle estimator MultinomialNB from version 0.19.2 when using version 0.20.3. This might lead to breaking code or invalid results. Use at your own risk.
  UserWarning)
/Users/gallaghe/Code/policytool/build/virtualenv/lib/python3.7/site-packages/sklearn/base.py:253: UserWarning: Trying to unpickle estimator Pipeline from version 0.19.2 when using version 0.20.3. This might lead to breaking code or invalid results. Use at your own risk.
  UserWarning)
/Users/gallaghe/Code/policytool/build/virtualenv/lib/python3.7/site-packages/numpy/matrixlib/defmatrix.py:71: PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
  return matrix(data, dtype=dtype, copy=False)
...........................
----------------------------------------------------------------------
Ran 27 tests in 0.136s

OK
```
# Checklist:

### DIFF
--- a/policytool/airflow/dags/policytool_dag.py
+++ b/policytool/airflow/dags/policytool_dag.py
@@ -9,6 +9,7 @@ import airflow.utils.dates
 
 from policytool.airflow.tasks.run_spiders_operator import RunSpiderOperator
 from policytool.airflow.tasks.parse_pdf_operator import ParsePdfOperator
+from policytool.airflow.tasks.extract_refs_operator import ExtractRefsOperator
 
 ORGANISATIONS = [
     'who_iris',
@@ -66,3 +67,30 @@ for organisation in ORGANISATIONS:
         dag=dag,
     )
     pdfParsing.set_upstream(run_spider)
+
+    parsed_pdf_file = os.path.join(
+        parsing_path,
+        'policytool-parse--parser-{organisation}.json'.format(
+            organisation=organisation
+        )
+    )
+
+    extracted_refs_path = os.path.join(
+        'datalabs-data',
+        'airflow',
+        'output',
+        'policytool-extract',
+        'test-extract-refs-{organisation}.json.gz'.format(
+            organisation=organisation
+        )
+    )
+
+    extract_refs = ExtractRefsOperator(
+        task_id='extract_refs',
+        model_path=parser_model,
+        src_s3_key=parsed_pdf_file,
+        dst_s3_key=extracted_refs_path,
+        dag=dag)
+
+    extract_refs.set_upstream(pdfParsing)
+
--- a/policytool/airflow/tasks/extract_refs_operator.py
+++ b/policytool/airflow/tasks/extract_refs_operator.py
@@ -59,10 +59,9 @@ class ExtractRefsOperator(BaseOperator):
                     model_path,
                     pool_map,
                     logger)
-                for doc_id, structured_references in refs:
+                for _, structured_references in refs:
                     for _, row in structured_references.iterrows():
                         item = dict(row)
-                        item['doc_id'] = doc_id
                         dst_f.write(json.dumps(item).encode('utf-8'))
                         dst_f.write(b'\n')
 
--- a/policytool/airflow/tasks/extract_refs_operator.py
+++ b/policytool/airflow/tasks/extract_refs_operator.py
@@ -16,8 +16,6 @@ from policytool.airflow.hook.wellcome_s3_hook import WellcomeS3Hook
 
 logger = logging.getLogger(__name__)
 
-RESOURCE_FILES = '/src/policytool/resources/'
-
 
 class ExtractRefsOperator(BaseOperator):
     """
@@ -62,14 +60,10 @@ class ExtractRefsOperator(BaseOperator):
                     pool_map,
                     logger)
                 for doc_id, structured_references in refs:
-                    for row in structured_references.iterrows():
-                        print(row)
-                        d = {'docId': doc_id}
-                        for (col_name, value) in enumerate(row):
-                            d.update({col_name : value})
-                        print(d)
-                        # TODO: update d with attrs from row (d.update()?)
-                        dst_f.write(json.dumps(d).encode('utf-8'))
+                    for _, row in structured_references.iterrows():
+                        item = dict(row)
+                        item['doc_id'] = doc_id
+                        dst_f.write(json.dumps(item).encode('utf-8'))
                         dst_f.write(b'\n')
 
             dst_rawf.flush()
--- a/policytool/airflow/tasks/extract_refs_operator.py
+++ b/policytool/airflow/tasks/extract_refs_operator.py
@@ -1,0 +1,83 @@
+"""
+Operator to run the web scraper on every organisation.
+"""
+import os
+import logging
+import tempfile
+import json
+import gzip
+
+from airflow.models import BaseOperator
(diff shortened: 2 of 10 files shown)


lizgzil:
this is ready for review, although i wasn't able to test the policytool_dag on the airflow UI (it only ever bring up the test_dag), not sure if this is as expected or not.

The test_dag works on my computer when I do and it has generated to json here : https://s3.console.aws.amazon.com/s3/buckets/datalabs-data/airflow/output/policytool-extract/?region=us-east-2&tab=overview

policytool/airflow/tasks/extract_refs_operator.py
```diff
@@ -0,0 +1,76 @@
...
+            path=self.dst_s3_key,
+        )
+        
+        pool_map = map
+
+        s3 = WellcomeS3Hook(aws_conn_id=self.aws_conn_id)
+
+        with tempfile.NamedTemporaryFile() as dst_rawf:
+            with gzip.GzipFile(mode='wb', fileobj=dst_rawf) as dst_f:
+                refs = yield_structured_references(
```
nsorros: @hblanks, i feel that yield_structured_references is misleading here. In this function we create the references and yield them. Yields functions have less functionality in my mind.

The main reason though I am bringing this up is that I also have a similarly named function that reads the files that is saved and then goes on to match the references so i wonder what should tield_structured_references do? read the file and yield the refererences or create the structure references and yield them

hblanks: Hmm. I'm not sure I follow the comment, so drop a note on slack or chat with me when you get a chance. We tend to use generators (functions with yield statements) in cases like this because we don't want to get all the references back at once -- we just want to take each one from wherever it came and write it to the output file.

policytool/airflow/dags/policytool_dag.py
```diff
@@ -66,3 +67,30 @@
         dag=dag,
     )
     pdfParsing.set_upstream(run_spider)
+
+    parsed_pdf_file = os.path.join(
+        parsing_path,
```
nsorros: shouldn't the parsing path be your input? instead of having to construct the parsed_pdf_file? parsing path should be the output of the previous step as far as i understand.

policytool/airflow/dags/policytool_dag.py
```diff
@@ -66,3 +67,30 @@
...
+        'policytool-parse--parser-{organisation}.json'.format(
+            organisation=organisation
+        )
+    )
+
+    extracted_refs_path = os.path.join(
```
nsorros: you might want to think whether you want to name this structured_references to keep the terminology consistent.

in terms of the task name i.e. ExtractRefsOperator, i think extract is a good verb to describe what we are doing. But you might want to differentiate between an Operator that would extract the reference strings and the parsed references using an intermediate word such as structured_references

policytool/airflow/dags/policytool_dag.py
```diff
@@ -66,3 +67,30 @@
...
+        'datalabs-data',
+        'airflow',
+        'output',
+        'policytool-extract',
+        'test-extract-refs-{organisation}.json.gz'.format(
+            organisation=organisation
+        )
+    )
+
+    extract_refs = ExtractRefsOperator(
```
nsorros: and similarly, this should be named refs i think. extract_refs is a better name for a function, if anything it should be called extracted_refs but i feel extracted is unnecessary.

policytool/airflow/tasks/extract_refs_operator.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class ExtractRefsOperator(BaseOperator):
+    """
+    Reads reference sections from a single newline-delimited
+    json.gz file, writing out parsed references as a single
+    newline-delimited json.gz file.
+
+    Args:
+        src_s3_key:
```
nsorros: needs to be filled

policytool/refparse/refparse.py
```diff
@@ -247,16 +200,74 @@ def parse_references(scraper_file, references_file, model_file,
...
+    )
+    matched_references_hard_text = text_matcher.hard_text_search(
+        doc
+    )
+    matched_references = remove_dups_and_concat(matched_references_parser, matched_references_hard_text)
+    return matched_references
+
+
+
+def match_references(scraper_file, references_file, model_file,
```
nsorros: extract_references or extract_structured_references is more consistent with the name of the airflow task

policytool/refparse/refparse.py
```diff
@@ -247,16 +200,74 @@ def parse_references(scraper_file, references_file, model_file,
...
+    fuzzy_matcher = FuzzyMatcher(
+        ref_file,
+        settings.FUZZYMATCH_THRESHOLD
+    )
+    text_matcher = HardTextSearch(ref_file)
+
+    refs = parse_references(
+        scraper_file, model_file, num_workers, logger)
+    for doc_id, structured_references in refs:
+        matched_references = run_match(
```
nsorros: intead of run_match this could be called match_references

policytool/refparse/refparse.py
```diff
@@ -247,16 +200,74 @@ def parse_references(scraper_file, references_file, model_file,
...
+    ref_file = get_file(references_file, 'csv')
+    check_references_file(ref_file, references_file)
+
+    fuzzy_matcher = FuzzyMatcher(
+        ref_file,
+        settings.FUZZYMATCH_THRESHOLD
+    )
+    text_matcher = HardTextSearch(ref_file)
+
+    refs = parse_references(
```
nsorros: instead of parse_references which is more of the name of one of the steps, the one that received a reference strings and returns structured references, this could be called extract_references

policytool/airflow/dags/policytool_dag.py
hblanks: Yeah, the confusion here is because these are task names...so they have to be lowercase (they're not classes, they're instances), and they probably should reflect the class name. Thus it seems OK to me.

PR #163, opened 2019-06-19
Changes to the match evaluation

lizgzil:
# Description

This PR:
- previously we used a csv I created a while ago as the evaluation data, but no code was uploaded to get this data yourself (and to get it I used the fuzzy matcher). This time we use a downloaded copy of the EPMC metadata file to make the evaluation data, which involves using the fuzzy matcher. This makes a few changes to the evaluation settings, importing the json not the csv in `evaluate_algo`, and big changes to `evaluate_match_references`.
- recall and precision are outputted from this evaluation now
- Ive updated the evaluation readme to explain this new method and included some figures from my exploratory work looking at the thresholds.
- Added `exploratory/investigate_match_thresholds.py` which allows someone to explore the cosine and the length thresholds, this should be able to be ran for different fuzzy match functions  depending on the changes made. Several plots are outputted from this.
- Cleaned up some of the exploration code - shortened lines and gave whole directories for relative imports.
- Added to the README to describe how to download the EPMC metadata

I would have added the length threshold to the fuzzy matcher but since Nick has changed the fuzzy matcher in the other PR I won't do this here.

## Type of change

- [ ] :sparkles: New feature

# How Has This Been Tested?

No tests, but this code is self contained anyway.


policytool/refparse/evaluate_algo.py
```diff
@@ -44,6 +46,33 @@ def yield_section_data(scrape_pdf_location, sections_location):
                 section_text = get_text('{}.txt'.format(section_path))
                 yield pdf_hash, section_name, section_text
 
+def load_pubs_json(pubs_file, total_N):
+
+    # ==== Load data to evaluate matching: ====
+
+    output_cols = ['title', 'pmid', 'pmcid']
+    with open(pubs_file, "r") as file:
```
nsorros: `with open(pubs_file, "r") as f:` f is possibly a better option since file is a reserved keyword

policytool/refparse/evaluate_algo.py
```diff
@@ -44,6 +46,33 @@ def yield_section_data(scrape_pdf_location, sections_location):
...
+        references = []
+        for line in range(total_N):
+            reference = json.loads(next(file))
```
nsorros:
```
for i, line in enumerate(f):
    if i >= total_N:
        break
    reference = json.loads(line)
```

policytool/refparse/evaluate_algo.py
```diff
@@ -44,6 +46,33 @@ def yield_section_data(scrape_pdf_location, sections_location):
...
+    evaluation_references['Document id'] = range(
+        0,
+        len(evaluation_references)
+    )
+    evaluation_references['Reference id'] = evaluation_references['pmid']
+    evaluation_references.rename(
+        index=str,
+        columns={'pmid': "uber_id"},
+        inplace = True
+    )
```
nsorros: A nice way of writing this that Hunter likes to use and I am trying to adopt is using yields. The benefit usually comes down to memory, even though in your case this does not apply the same way since you need to "materialise" the data instead of process them as they come (one by one) but even so you avoid creating two variables that contain the data i.e. evaluation_references and references this way. The code would look something like this:
```
for i, line in enumerate(f):
    if i >= total_N:
        return
    reference = json.loads(line)
    if all([output_col in reference for output_col in output_cols]):
        yield {
            'Reference id':  reference['pmid'],
            'Document id': i,
            'title': title
    }
```

nsorros:
This is something we need to involve @aoifespenge to. @aoifespenge let us know what is the best way to discuss the choice of thresholds, we can also do it in a separate issue.

policytool/refparse/README.md
```diff
@@ -59,3 +59,26 @@ python ./parse_latest.py msf \
...
+brew install poppler
+```
+and then downloading the evaluation data from [here](https://s3-eu-west-1.amazonaws.com/datalabs-data/policy_tool_tests) and storing it in `algo_evaluation/data_evaluate/`, which can be done in the command line by running
+```
+aws s3 cp --recursive s3://datalabs-data/policy_tool_tests algo_evaluation/data_evaluate/
+```
+and
+```
+aws s3 cp s3://datalabs-staging/airflow/output/open-research/epmc-metadata/epmc-metadata.json.gz .
+download: s3://datalabs-staging/airflow/output/open-research/epmc-metadata/epmc-metadata.json.gz to algo_evaluation/data_evaluate/epmc-metadata.json.gz
```
nsorros: this line is redundant i think

policytool/refparse/README.md
```diff
@@ -59,3 +59,26 @@ python ./parse_latest.py msf \
...
+gunzip epmc-metadata.json.gz
+```
+and finally running
+```
+python evaluate_algo.py --verbose True
```
nsorros: i thought we removed verbose true

policytool/refparse/README.md
```diff
@@ -59,3 +59,26 @@ python ./parse_latest.py msf \
...
 
+## Evaluating each component of the algorithm
+
+We have devised some evaluation data in order to evaluate 5 steps of the model. The results can be calculated by first installing poppler
+```
+brew install poppler
+```
+and then downloading the evaluation data from [here](https://s3-eu-west-1.amazonaws.com/datalabs-data/policy_tool_tests) and storing it in `algo_evaluation/data_evaluate/`, which can be done in the command line by running
+```
+aws s3 cp --recursive s3://datalabs-data/policy_tool_tests algo_evaluation/data_evaluate/
```
nsorros: can we make this data public by any chance @hblanks ?

policytool/refparse/algo_evaluation/evaluation.md
```diff
@@ -52,15 +52,38 @@ abs(100*((predicted number - actual number) / actual number))
...
-- We use `FuzzyMatcher` from `policytool/refparse/utils` to predict whether the reference titles in the evaluation data are matched with any of those from the publication set.
-- We compare 'Do we expect a match?' with whether we found a true match or not, returning the F1 score.
-- We also return a classification report and the frequency table of match types (the reference was matched to the same reference in the publication set (True), the reference was matched to a different reference in the publication set (False), or there was no match found).
+- We predict whether a match was found (predicted = 'Positive') or not (predicted = 'Negative') by asking was the cosine similarity over a threshold and was the title length over a threshold for each of the 20,000 matches.
+- We find the F1 score of these actual and predicted lists.
+- We also record whether the match found was correct or not, which can be found by comparing the uber ids of the reference and it's match. Thus we also return a classification report and the frequency table of match types.
 
+### Evaluation Thresholds
```
nsorros: @aoifespenge here is the point that needs discussion

policytool/refparse/algo_evaluation/evaluation.md
```diff
@@ -52,15 +52,38 @@ abs(100*((predicted number - actual number) / actual number))
...
+For this evaluation we also considered the thresholds to use when predicting whether a match should be taken forward or not. For this we plotted all the cosine similarities found from the actual = "Negative" set (i.e. the second best matches). We also investigated the relationship between cosine similarity and title length.
 
+<p float="left">
+  <img src="exploratory/negative_cosines_hist_2019-06-19-1335.png" width="500" />
+  <img src="exploratory/negative_cosines_len_scatter_2019-06-19-1335.png" width="250" />
+</p>
+
+From these we can see that generally the cosine similarity is quite low, but when it is high (>0.8) the title length tends to be quite short. Thus we saw that if we set the match and title length thresholds to be relatively high then we'd reduce the number of false positives. The 95th percentile of the cosine similarities is 0.6 and the 5th percentile of the title length is 33.
```
nsorros: This suggest to use 0.6 and 33 as the thresholds right? What is the 99th percentile and 1th percentile?

policytool/refparse/README.md
lizgzil: that was in the other PR that was merged

policytool/refparse/evaluate_algo.py
```diff
@@ -4,17 +4,19 @@
...
-from algo_evaluation.evaluate_find_section import evaluate_find_section
-from algo_evaluation.evaluate_split_section import evaluate_split_section
-from algo_evaluation.evaluate_parse import evaluate_parse
-from algo_evaluation.evaluate_match_references import evaluate_match_references
+from policytool.refparse.utils import FileManager
+from policytool.refparse.algo_evaluation.evaluate_settings import settings
+from policytool.refparse.algo_evaluation.evaluate_find_section import evaluate_find_section
+from policytool.refparse.algo_evaluation.evaluate_split_section import evaluate_split_section
+from policytool.refparse.algo_evaluation.evaluate_parse import evaluate_parse
+from policytool.refparse.algo_evaluation.evaluate_match_references import evaluate_match_references
```
nsorros: @hblanks is that how we should import or we should run this using the module flag? `python -m policytool.refparse`

policytool/refparse/algo_evaluation/evaluation.md
```diff
@@ -52,15 +52,38 @@ abs(100*((predicted number - actual number) / actual number))
...
-- The positive list was reduced to 199 since one reference didn't have a second strong match, and the negative list gave 202 references since for some references there were multiple references with the same cosine similarity competing for second place.
-- We take the titles and publication IDs from these and combine them in "positive_and_negative_match_test_data.csv", and in this create a 'Do we expect a match?' column where we record a '1' if this reference was from the positive list and a '0' if it was from the negative list.
+- We selected the first 100,000 references from a list of EPMC publications ("epmc-metadata.json").
+- We randomly selected a sample 10,000 from these.
+- Each of the references from this sample were matched (using `FuzzyMatcher` from `policytool/refparse/utils`) against both the original 100,000 references, and against the original list with the sample list removed (i.e. 90,000 references). No thresholds were used in these matches. We record the title length and the cosine similarity during these matchings.
```
aoifespenge: Can you explain what the effect of not using thresholds here is?

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,87 @@
 import pandas as pd
 
-from sklearn.metrics import classification_report, f1_score
+from sklearn.metrics import classification_report, f1_score, recall_score, precision_score
 
-from utils import FuzzyMatcher
+from policytool.refparse.utils import FuzzyMatcher
+
+
+def get_match_data(evaluation_references, length_threshold, sample_N):
```
nsorros: I don't think you use length_threshold, if not you can remove it

policytool/refparse/algo_evaluation/evaluation.md
aoifespenge: I found this difficult to get my head around. Would it be possible write in a few lines of what are the implications of this.... what are the takeaways of this finding?

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,87 @@
...
+    return prediction
 
 def evaluate_metric(actual, predicted):
 
-    actual_binary = actual.astype(int)
-    # predicted is True, False and "No match found",
-    # convert to 1 (True) or 0 (False or "No match found")
-    predicted_binary = [1 if pred_type==True else 0 for pred_type in predicted]
+    actual_binary = ["Positive" if "Positive" in actual_type else "Negative" for actual_type in actual]
+    predicted_binary = ["Positive" if "Positive" in pred_type else "Negative" for pred_type in predicted]
```
nsorros: It would be preferable if this was not needed. ideally the actual and predicted vectors are in the format that f1_score, recall_score etc wants them which is arrays or list of 0 and 1

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,87 @@
...
+
+    matched_publication_positive['Match Type'] = ["Positive"]*sample_N
+    matched_publication_negative['Match Type'] = ["Negative"]*sample_N
+
+    eval_references = pd.concat([matched_publication_positive, matched_publication_negative])
+    eval_references = eval_references[['Title', 'title', 'Cosine_Similarity', 'Match Type', 'Title Length', 'uber_id','Reference id']]
+
+    return eval_references
+
+def predict_match(match, match_threshold, length_threshold):
```
nsorros: I would pass the matcher here instead of the match. The matcher should be initialised with all publications apart from the negative ones. I would also pass the match_data in terms of titles and and lengths.

`predict_match_data(matcher, match_data, match_threshold. length_threshold)`

This is more in line with how you would create a function given an ML model which would be something like

`predict_match_data(model, X, match_threshold, length_threshold)`

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,87 @@
...
+    matched_publication_positive['Title Length'] = [len(title) for title in matched_publication_positive['Title']]
+    matched_publication_negative['Title Length'] = [len(title) for title in matched_publication_negative['Title']]
+
+    matched_publication_positive['Match Type'] = ["Positive"]*sample_N
+    matched_publication_negative['Match Type'] = ["Negative"]*sample_N
+
+    eval_references = pd.concat([matched_publication_positive, matched_publication_negative])
+    eval_references = eval_references[['Title', 'title', 'Cosine_Similarity', 'Match Type', 'Title Length', 'uber_id','Reference id']]
+
+    return eval_references
```
nsorros: here i would probably return two things
* match_data which would be three columns title, title_length, match
* matcher initialised in all references but the negatives or all the references but the negatives ones or a custom matcher class with a predict function that you give a single match data item and returns whether it matches or not

policytool/refparse/algo_evaluation/evaluation.md
nsorros: This is slightly confusing but here Liz just describes how she created the data not how she "predicted" a match where the threshold comes into play more.

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -0,0 +1,220 @@
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+import seaborn as sns
+from datetime import datetime
```
nsorros: separate python libraries from external libraries with a space

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -0,0 +1,220 @@
...
+def get_predict_result(match, cosine_threshold, length_threshold):
+
+    if ( 
+        (
+            match['Cosine_Similarity'] > cosine_threshold
+        ) and 
+        (
+            match['Title Length'] > length_threshold
+        )
+        ):
```
nsorros: I think one parenthesis is enough

```
if (
    match['Cosine_Similarity'] > cosine_threshold and
    match['Title Length'] > length_threshold
):

File: policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py [hunk 17]
```diff
@@ -0,0 +1,220 @@
...
+    return prediction
+
+
+def make_neg_pos_plots(
+        matched_pubs_negative,
+        matched_pubs_positive,
+        incorrect_matched_pubs_positive,
+        N,
+        now
+        ):
```
nsorros: I am not sure that is more readable, i think it is better to stack them in multiple of threes or twos

```
def make_neg_pos_plots(
    matched_pubs_negative, matched_pubs_positive, 
    incorrect_matched_pubs_positive, N, now):
```

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -0,0 +1,220 @@
...
+        round(perc, 2)
+        )
+    )
+
+if __name__ == '__main__':
+
+    EVAL_PUB_DATA_FILE_NAME = "epmc-metadata.json"
+    FOLDER_PREFIX = "../data_evaluate"
+
+    empc_file = os.path.join(FOLDER_PREFIX, EVAL_PUB_DATA_FILE_NAME)
```
nsorros: it would be better if this path was relative to this file
`os.path.join(os.path.dirname(__file__), FOLDER_PREFIX, PUB_DATA_FILE_NAME)`

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -0,0 +1,220 @@
...
+    )
+
+    matched_pubs_negative = eval_references[
+        eval_references['Match Type']=="Negative"
+    ]
+    matched_pubs_positive = eval_references[
+        (
+            (eval_references['Match Type']=="Positive") &
+            (eval_references['uber_id']==eval_references['Reference id'])
+        )
```
nsorros: this external parenthesis is not needed i think

policytool/refparse/algo_evaluation/evaluation.md
aoifespenge: As discussed, it would be great to have that higher level sentence to explain the implication: e.g. " The 95th percentile of the cosine similarities is 0.6 and the 5th percentile of the title length is 33. Therefore I can get rid of 95% of .... by ...... "

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -0,0 +1,220 @@
...
+    eval_references = get_match_data(
+        evaluation_references,
+        length_threshold,
+        sample_N
+    )
+
+    matched_pubs_negative = eval_references[
+        eval_references['Match Type']=="Negative"
+    ]
+    matched_pubs_positive = eval_references[
```
nsorros: this should be named true_positives or true_positive_matches. try to use common terminology when possible, it would make reviews and future development easier

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -0,0 +1,220 @@
...
+        (
+            (eval_references['Match Type']=="Positive") &
+            (eval_references['uber_id']==eval_references['Reference id'])
+        )
+    ]
+    incorrect_matched_pubs_positive = eval_references[
```
nsorros: false_positives

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -0,0 +1,220 @@
...
+            (eval_references['Match Type']=="Positive") &
+            (eval_references['uber_id']!=eval_references['Reference id'])
+        )
+    ]
+
+    now = datetime.now()
+    make_neg_pos_plots(
+        matched_pubs_negative,
+        matched_pubs_positive,
+        incorrect_matched_pubs_positive,
```
nsorros: i think you should pass eval_references here and calculate true and false positives as well as true negatives inside the function.

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -0,0 +1,220 @@
...
+    match_scores = []
+    for cosine_threshold in np.linspace(0.5, 1, 10, endpoint=True):
+        for length_threshold in  np.arange(0, 150, 5):
+            predictions = [
+                get_predict_result(
+                    match, cosine_threshold, length_threshold
+                ) for i, match in eval_references.iterrows()
+            ]
+            f1 = round(f1_score(
+                eval_references['Match Type'].to_list(),
```
nsorros: do you need to do to_list every time? if so, you can save this in a variable as predictions or match_pred or something

aoifespenge:
@lizgzil that was a great conversation we had around the implications of this exercise. I'll summarise it here:
A lot of the analysis here is in specific data science, which it needs to remain concrete, and there is a risk in abstracting it for the layperson as it will lose meaning. However, for someone who isn't involved in the data science process, it is too specific and lay people cannot draw meaning from it. As a remedy we suggested that having a separate summary section that does explain things at an abstracted level. By it being separate it doesn't mix in with the technical language and take away its specificity. These summary sections would cover: What is at stake? What are the trade-offs? What are the implications?  e.g. in this case, it is about deciding:
Do we want to be confident that matched are correct but we also false negatives?
Do we want to be confident that we haven’t got any false positives at the expense of missing some true positives?

I think just highlighting this it would spark discussion amongst our team about what the impact of these choices would be. For example, if in the case of risking excluding false negatives (missing things) then that means that we are likely to titles that are short and generic.... What if short and/or generic titles are associated with certain publication characteristics such as university, gender of author, discipline, topic etc..... Of course, we have no idea if there is any link and the only way to find out would by carrying out social research.... So we would need to make a call on whether we want to do that research.

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -0,0 +1,220 @@
...
+
+from sklearn.metrics import (
+    f1_score, recall_score, precision_score
+    )
+from policytool.refparse.algo_evaluation.evaluate_match_references\
+    import get_match_data
+
+from policytool.refparse.evaluate_algo import load_pubs_json
+
+def get_predict_result(match, cosine_threshold, length_threshold):
```
nsorros: this function can be refactored and combined with `predict_match`

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -0,0 +1,220 @@
...
+    total_N = 100000
+    sample_N = 10000
+
+    print("===== Loading {} lines of the EMPC data".format(total_N),
+        "and getting match evaluation data for",
+        "a random {} of these =====".format(sample_N)
+        )
+    evaluation_references = load_pubs_json(empc_file, total_N)
+    length_threshold = -1 # Effectively no threshold
+    eval_references = get_match_data(
```
nsorros: the way this code is written right now, if we want to investigate the threshold for a different matcher we need to change evaluate_match_references which is weird. it would be better to design this as suggested above which is that this function returns a matcher and a dataset

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -0,0 +1,220 @@
...
+            match_scores.append(
+                {'Match threshold' : cosine_threshold,
+                'Length threshold' : length_threshold,
+                'F1 Score' : f1,
+                'Recall' : recall,
+                'Precision' : precision
+                })
+        print(cosine_threshold)
+
+    match_scores_df = pd.DataFrame(match_scores).round(4)
```
nsorros: i would do that inside the function make_heat_plots under the justification that you "transform" the data into another data structure that is only relevant to that function so the outside world does not need to know

policytool/refparse/README.md
```diff
@@ -59,3 +59,26 @@ python ./parse_latest.py msf \
...
+```
+and
+```
+aws s3 cp s3://datalabs-staging/airflow/output/open-research/epmc-metadata/epmc-metadata.json.gz .
```
nsorros: i think this should go to data_evaluate

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,87 @@
...
+
+    evaluation_references_negative = evaluation_references.loc[~evaluation_references['uber_id'].isin(references_sample['uber_id'])]
+    
+    fuzzy_matcher_positive = FuzzyMatcher(evaluation_references, match_threshold)
+    fuzzy_matcher_negative = FuzzyMatcher(evaluation_references_negative, match_threshold)
+
+    matched_publication_positive = fuzzy_matcher_positive.match_vectorised(references_sample)
+    matched_publication_negative = fuzzy_matcher_negative.match_vectorised(references_sample)
+
+    matched_publication_positive['Title Length'] = [len(title) for title in matched_publication_positive['Title']]
```
nsorros: `matched_publication_positive['Title'].apply(lambda x: len(x))` should be quicker

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,87 @@
...
+
+
+def get_match_data(evaluation_references, length_threshold, sample_N):
+
+    # To get these you want all the matches, regardless of cosine
+    match_threshold = -1 
+
+    # Take a random sample of the evaluation references to find matches for
+    references_sample = evaluation_references.sample(n = sample_N, random_state = 0)
+    references_sample['Title'] = references_sample['title']
```
nsorros: why are you copying this field?

nsorros:
You need to include matplotlib in the requirements

policytool/refparse/algo_evaluation/evaluation.md
lizgzil: 99th percentile for the cosine similarities of the negative set = 0.91
1th percentile for the title lengths of the correctly matched positive set = 19.0

policytool/refparse/README.md
hblanks: So long as it's not fed by dimensions, we can. We'll want to create a dedicated public bucket for it. `s3://datalabs-public` or such. LMK if you'd like a hand setting that up, or if you want to do that as part of the pre-release sweep.

policytool/refparse/evaluate_algo.py
hblanks: You can either use the full path (as done in this change, `from policytool.refparse....`), or you should use relative paths:

```
from .algo_evaluation.evaluate_parse import evaluate_parse
```

Both are fine. But as noted, relative paths mean you're gonna have to do `python -m policytool.refparse.evaluate_algo` instead of `python ./evaluate_algo.py`.

policytool/refparse/evaluate_algo.py
```diff
@@ -128,21 +158,20 @@ def create_argparser():
...
-    )
 
-    # Load WT publications to match references against
-    logger.info('[+] Reading {}'.format(settings.MATCH_PUB_DATA_FILE_NAME))
-    publications = fm.get_file(
-        settings.MATCH_PUB_DATA_FILE_NAME,
+    logger.info('[+] Reading the first {} lines of {}'.format(
+        settings.EVAL_MATCH_NUMBER,
+        settings.EVAL_PUB_DATA_FILE_NAME
+        ))
```
hblanks: One note is that the logger actually has built-in `%s` formatting. So, this could be:

```python
logger.info(
    'main: reading the first %d lines of %s',
    settings.EVAL_MATCH_NUMBER,
    settings.EVAL_PUB_DATA_FILE_NAME
)
```

Note I also tend to avoid punctional prefixes in log lines in favor of function names or main as above.

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
lizgzil: this doesnt work, but `os.path.join(os.path.dirname('__main__'), FOLDER_PREFIX, PUB_DATA_FILE_NAME)` does. Is that ok?

policytool/refparse/evaluate_algo.py
hblanks: Oh, and love seeing some logging in here!

policytool/refparse/README.md
lizgzil: it does seem a shame we can't just download all the evaluation data from one location, but if we copied this file into data_evaluation then we'd have to make sure we were keeping it up to date, unless you'd be happy to just keep a copy of an old version in the data_evaluate folder?

policytool/refparse/algo_evaluation/evaluate_match_references.py
lizgzil: we need a field called 'Title' and one called 'title' for the fuzzy matcher code. Originally 'Title' was from the output of the parser and 'title' was from the EPMC data (or the other way around?), but in this case they are just the same data. I could open an issue to make the fuzzymatcher take as an input the column names for these?

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
lizgzil: i've restructured `evaluate_match_references` and `investigate_match_thresholds.py` in such a way that I no longer have 'get_match_data'

policytool/refparse/algo_evaluation/evaluate_match_references.py
lizgzil: I want `evaluate_metric` to output both the F1 etc (which would indeed have the binary input), but also the frequency table of all the types of matches, which requires the none binary input. This gives us extra information about the different types of matches:
```
Matched correctly?     False  True  No match found
Do we expect a match?                             
0                         15     0             187
1                          1   198               0
```
thus this gives us the additional info that there are 15 times where we weren't expecting a match but got one, and the 1 time when we were expecting a match, but got the wrong one.

lizgzil:
I think I've addressed most of the comments by  @nsorros @hblanks and @aoifespenge (either in Github replies or in the commit).

FYI about my change to `evaluate_find_section`:
I found that on 2 occasions the filenames of the pdfs in the pdfs folder were being read with their '.pdf' extension. Strange since I wasn't aware that I edited anything to cause this change!
You will see that I've fixed it by :
```
if os.path.splitext(pdf_name)[1] == ".pdf":
    pdf_name = os.path.splitext(pdf_name)[0]
```
The problematic files were '85467b4370567191f4ff9c9509feaadc.pdf' and 'ddce9b7fdfa47637e298a1a2e77c715d.pdf', not sure why they are any different from the other pdfs in that folder though, they are all named with the extension.

policytool/refparse/algo_evaluation/evaluate_find_section.py
```diff
@@ -147,6 +146,9 @@ def scrape_process_pdf(
         scrape_data : a list of dicts with the predicted and actual texts for
             each of the sections we looked for in the pdf
     """
+    if os.path.splitext(pdf_name)[1] == ".pdf":
+        pdf_name = os.path.splitext(pdf_name)[0]
```
nsorros: i did not know of splittext, nice find 👍

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -0,0 +1,235 @@
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+import seaborn as sns
+
+from datetime import datetime
+import os
+
+from sklearn.metrics import (
```
nsorros: did hunter mention to separate the from imports from the rest? i usually keep them together in which case this would be grouped with the firs set

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -0,0 +1,235 @@
...
+    f1_score, recall_score, precision_score
+    )
+
+from policytool.refparse.evaluate_algo import load_pubs_json
+
```
nsorros: no need for the blank line

policytool/refparse/evaluate_algo.py
nsorros: @lizgzil is that clear? if not happy to discuss it more

unpinned_requirements.txt
```diff
@@ -6,6 +6,7 @@ celery
 editdistance
 elasticsearch
 falcon
+matplotlib
```
nsorros: I am under the impression that when you add a requirement here you need to run `make update-requirements-txt` which will also update the requirements.txt. @hblanks can confirm. maybe we can add some instructions in the readme if there are none already.

policytool/refparse/evaluate_algo.py
```diff
@@ -4,17 +4,19 @@
 
 from argparse import ArgumentParser
 import os
+import json
+import pandas as pd
```
nsorros: this dependency is not from the standard library

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
SamDepardieu: Just saw that in my notifications, here's the deal:
https://www.python.org/dev/peps/pep-0008/#imports

Precisely:
```
Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

Imports should be grouped in the following order:

    Standard library imports.
    Related third party imports.
    Local application/library specific imports.

You should put a blank line between each group of imports.
```

policytool/refparse/evaluate_algo.py
```diff
@@ -44,6 +46,25 @@ def yield_section_data(scrape_pdf_location, sections_location):
                 section_text = get_text('{}.txt'.format(section_path))
                 yield pdf_hash, section_name, section_text
 
+def load_pubs_json(pubs_file, total_N):
```
nsorros: this should better be named yield_pubs_json as load make it seem like it gives a data structure back

policytool/refparse/evaluate_algo.py
```diff
@@ -120,65 +141,66 @@ def create_argparser():
...
-        'csv'
-    )
-
+        settings.EVAL_PUB_DATA_FILE_NAME
+        )
+    evaluation_references = load_pubs_json(
+        pubs_file,
+        settings.EVAL_MATCH_NUMBER
+        )
+    evaluation_references = pd.DataFrame(evaluation_references)
```
nsorros: and here actually I am surprised that this works since load_pubs_json now yields one item at a time, it might have to do with dataframe been able to work with an iterable but it would be better to instantiate the items into a list at this point.
```
evaluation_references = [p for p in yield_pubs_json(..)]
```
otherwise name it like a generator
```
evaluation_references_generator = yield_pubs_json(..)
```

nsorros:
@lizgzil have you addressed these comments?
* change predict_match so that it accepts a matcher and two columns from match_data (title and uber_id) along with match_threshold and length_threshold and returns predictions
* change evaluate_metric to work with binary vectors of actual and predicted, essentially the predictions from predict_match and the third column from match_data

policytool/refparse/evaluate_algo.py
lizgzil: yes, but I thought I'd just keep my original since hunter said either way was fine. Would you prefer me to do the relative paths?

lizgzil:
> @lizgzil have you addressed these comments?
> 
> * change predict_match so that it accepts a matcher and two columns from match_data (title and uber_id) along with match_threshold and length_threshold and returns predictions

`predict_match_data` is almost what you want, but I passed all of match_data, not just the 2 columns, but i can change that

lizgzil:
> * change evaluate_metric to work with binary vectors of actual and predicted, essentially the predictions from predict_match and the third column from match_data

This hasn't been done, but I will mention you in the comment in which I describe why not

policytool/refparse/algo_evaluation/evaluate_match_references.py
lizgzil: @nsorros this is where I describe why I didnt use a binary input

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,95 @@
...
+
+    predict_match_data_partial = partial(
+        predict_match_data,
+        match_data=match_data,
+        match_threshold=match_threshold,
+        length_threshold=length_threshold
+        )
+
+    predictions_positive = predict_match_data_partial(fuzzy_matcher_positive)
+    predictions_negative = predict_match_data_partial(fuzzy_matcher_negative)
```
nsorros: can we not have one FuzzyMatcher? instantiated with all publications but the negative ones?

nsorros:
> > * change evaluate_metric to work with binary vectors of actual and predicted, essentially the predictions from predict_match and the third column from match_data
> 
> This hasn't been done, but I will mention you in the comment in which I describe why not

It seems to me that maybe what you need is to return the match['uber_id'] as predictions to be able to calculate what you want.

policytool/refparse/algo_evaluation/evaluate_match_references.py
nsorros: i think this problem will be solved if actual and predicted are the uber_id and reference_id

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,95 @@
...
+            match['Cosine_Similarity'] > match_threshold and 
+            len(match['Title']) > length_threshold
+        ):
+            if match['uber_id'] == match['Reference id']:
+                predictions.append("Positive - correct match")
+            else:
+                predictions.append("Positive - incorrect match")
+        else:
+            if match['uber_id'] == match['Reference id']:
+                predictions.append("Negative - correct match")
```
nsorros: I am bit confused as to what this means, false negative?

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,95 @@
...
+            else:
+                predictions.append("Negative - incorrect match")
+
+    return predictions
```
nsorros: maybe predictions should be the id of the match so that you can calculate more detailed metrics

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,95 @@
...
+
+
+def predict_match_data(matcher, match_data, match_threshold, length_threshold):
+
+    matched_publications = matcher.match_vectorised(match_data)
+
+    predictions = []
+    for _, match in matched_publications.iterrows():
+        if (
+            match['Cosine_Similarity'] > match_threshold and
```
nsorros: also match_vectorised already checks for the threshold, why do we do it again?

policytool/refparse/algo_evaluation/evaluate_match_references.py
nsorros: cause they way it is frased it seems like a true negative but from the code i think it is a false negative

nsorros:
And when the code is ready to merge we need to make a decision on the thresholds based on the investigation and after consulting @aoifespenge

policytool/refparse/algo_evaluation/evaluate_match_references.py
lizgzil: hm but this doesnt give any information about whether it passed the thresholds or not, not neccessarily that the uber_id and reference_id are the same. Also it also gives no information about whether we expected to find a match or not.

policytool/refparse/algo_evaluation/evaluation.md
```diff
@@ -52,15 +52,43 @@ abs(100*((predicted number - actual number) / actual number))
...
-- We compare 'Do we expect a match?' with whether we found a true match or not, returning the F1 score.
-- We also return a classification report and the frequency table of match types (the reference was matched to the same reference in the publication set (True), the reference was matched to a different reference in the publication set (False), or there was no match found).
+- We predict whether a match was found (predicted = 'Positive') or not (predicted = 'Negative') by asking was the cosine similarity over a threshold and was the title length over a threshold for each of the 20,000 matches.
+- We find the F1 score of these actual and predicted lists.
+- We also record whether the match found was correct or not, which can be found by comparing the uber ids of the reference and it's match. Thus we also return a classification report and the frequency table of match types.
 
+### Evaluation Thresholds
 
+Summary: Filtering out 'matches' where the cosine similarity is less than 0.6 and the title length is less than 33 will remove lots of false positives. However, this comes at the expensive of filtering out some true positives.
```
aoifespenge: Nice. Could you make it even more explicit that a decision has been made:

We decided to prioritise making the matched data more accurate (removing false positives) at the expense of missing some true positives.  

What is the impact of this? Some references that should be matched are will be missed. Specifically, from the way we set the thresholds, references with short titles and those that are generic will be more likely to be excluded. The next question is whether this is likely like to have any unfair outcomes for the users and communities our product affects? To be discussed in this issue: https://github.com/wellcometrust/datalabs/issues/352

nsorros: We did not decide to make it more accurate, as missing true positives which means increasing false negatives is still a penalty. We decided to optimize for precision is the best term. Precision quantifies the number of false positives over predicted positives. 

This means that we are excluding some matches that would be correct and as you said we need to investigate whether that has an effect yes. But I would say you can chip in even in this decision, which is to maximise precision. Maybe you think this is the wrong thing to do.

lizgzil:
@nsorros I've made the changes we talked about in the latest commit, see what you think

policytool/refparse/algo_evaluation/evaluation.md
lizgzil: i've added 'Thus, we made a decision (which can be challenged) that we would optimise for precision, which means that we want to be confident that our predicted matches are correct, whilst knowing that we won't have predicted all of the matches.', what do you think?

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
```diff
@@ -10,30 +10,31 @@
...
+    evaluation_references_without_negative = evaluation_references.loc[
+        ~evaluation_references['uber_id'].isin(match_data_negative['uber_id'])
         ]
-    
-    fuzzy_matcher_positive = FuzzyMatcher(evaluation_references, -1)
-    fuzzy_matcher_negative = FuzzyMatcher(evaluation_references_negative, -1)
 
-    matched_publications_positive = fuzzy_matcher_positive.match_vectorised(match_data)
-    matched_publications_negative = fuzzy_matcher_negative.match_vectorised(match_data)
+    fuzzy_matcher = FuzzyMatcher(evaluation_references_without_negative, -1)
```
nsorros: did you keep the -1 threshold?

policytool/refparse/evaluate_algo.py
```diff
@@ -5,12 +5,13 @@
 from argparse import ArgumentParser
 import os
 import json
-import pandas as pd
 from os import listdir 
 from datetime import datetime
 from urllib.parse import urlparse
 from collections import defaultdict
 
+import pandas as pd
```
nsorros: is that used?

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,85 @@
...
+    for (a,p) in zip(actual, predicted):
+        if p:
+            if a==p:
+                predicted_types.append("Correct match found")
+            else:
+                predicted_types.append("Incorrect match found")
+        else:
+            predicted_types.append("No match found")
+
+    df_pairs = pd.DataFrame(
```
nsorros: df_pairs is not a good name

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,85 @@
...
+from collections import Counter
+
 import pandas as pd
+from sklearn.metrics import classification_report, f1_score, recall_score, precision_score
 
-from sklearn.metrics import classification_report, f1_score
+from policytool.refparse.utils import FuzzyMatcher
 
-from utils import FuzzyMatcher
+def predict_match_data(matcher, match_data):
```
nsorros: it would be nice to add some doclines describing the input and output

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,85 @@
...
+        predictions.append(matched_publications['uber_id'][0])
 
-    actual_binary = actual.astype(int)
-    # predicted is True, False and "No match found",
-    # convert to 1 (True) or 0 (False or "No match found")
-    predicted_binary = [1 if pred_type==True else 0 for pred_type in predicted]
+    return predictions
 
-    similarity = round(f1_score(actual_binary, predicted_binary, average='micro'), 3)
+def evaluate_metric(actual, predicted):
```
nsorros: same here, it would be nice to have some documentation about what actual and predicted is, especially now that this is not binary which is the most usual case

policytool/refparse/algo_evaluation/evaluate_match_references.py
```diff
@@ -1,53 +1,85 @@
...
 
+    predicted_bin = [a==p for (a,p) in zip(actual, predicted)]
```
nsorros: also a comment here describing briefly why we do this, reminding your future self that two ids need to be the same for the match to be correct

policytool/refparse/algo_evaluation/evaluate_match_references.py
nsorros: also small stylistic change it would be nice to remove the newline from above and add it after since actual_types, predicted_types and df_pairs are related with each other and the block is calculating the metrics

policytool/refparse/algo_evaluation/evaluation.md
```diff
@@ -52,15 +52,43 @@ abs(100*((predicted number - actual number) / actual number))
...
-- The positive list was reduced to 199 since one reference didn't have a second strong match, and the negative list gave 202 references since for some references there were multiple references with the same cosine similarity competing for second place.
-- We take the titles and publication IDs from these and combine them in "positive_and_negative_match_test_data.csv", and in this create a 'Do we expect a match?' column where we record a '1' if this reference was from the positive list and a '0' if it was from the negative list.
+- We selected the first 100,000 references from a list of EPMC publications ("epmc-metadata.json").
+- We randomly selected a sample 10,000 from these.
+- Each of the references from this sample were matched (using `FuzzyMatcher` from `policytool/refparse/utils`) against both the original 100,000 references, and against the original list with the sample list removed (i.e. 90,000 references). No thresholds were used in these matches. We record the title length and the cosine similarity during these matchings.
```
nsorros: Note that this might be inaccurate now since we use 1000 and the default threshold, hopefully

policytool/refparse/evaluate_algo.py
nsorros: i just thought you experienced problems running it so wanted to remind you how relatives paths are supposed to be run. also remind myself :p

policytool/refparse/evaluate_algo.py
```diff
@@ -44,6 +47,25 @@ def yield_section_data(scrape_pdf_location, sections_location):
...
+    output_cols = ['title', 'pmid']
+    with open(pubs_file, "r") as f:
+        references = []
+        for i, line in enumerate(f):
+            if i >= total_N:
+                break
+            reference = json.loads(line)
+            if all([output_col in reference for output_col in output_cols]):
+                yield {
+                'Document id': i,
```
nsorros: a tab is missing here

policytool/refparse/evaluate_algo.py
```diff
@@ -120,65 +142,66 @@ def create_argparser():
...
         settings.MODEL_FILE_PREFIX,
         settings.MODEL_FILE_TYPE
     )
 
-    # ==== Get the evaluation metrics ====
+    # # ==== Get the evaluation metrics ====
     logger.info('\nStarting the evaluations...\n')
 
-    logger.info('[+] Running evaluations 1 and 2')                     
+    logger.info('main: Running evaluations 1 and 2')
```
nsorros: When you tidy the code it would be nice to give more descriptive names to the tests

policytool/refparse/algo_evaluation/evaluation.md
```diff
@@ -52,15 +52,43 @@ abs(100*((predicted number - actual number) / actual number))
...
+- Thus we have a list of 20,000 reference matches, 10,000 where they should match exactly (actual = 'Positive'), and 10,000 where the second most similar reference should be matched (actual = 'Negative').
 
 ### Evaluation Score 5
-- We remove the references with 'Do we expect a match?' = 0 from "uber_api_publications.csv", this is our 'publication set'.
-- We use `FuzzyMatcher` from `policytool/refparse/utils` to predict whether the reference titles in the evaluation data are matched with any of those from the publication set.
-- We compare 'Do we expect a match?' with whether we found a true match or not, returning the F1 score.
-- We also return a classification report and the frequency table of match types (the reference was matched to the same reference in the publication set (True), the reference was matched to a different reference in the publication set (False), or there was no match found).
+- We predict whether a match was found (predicted = 'Positive') or not (predicted = 'Negative') by asking was the cosine similarity over a threshold and was the title length over a threshold for each of the 20,000 matches.
```
nsorros: slightly inaccurate in light of latest changes.

policytool/refparse/evaluate_algo.py
lizgzil: it is used on line 166 to convert the input to `evaluate_match_references` to a dataframe. (I use some pandas functions in `evaluate_match_references` so it is helpful that the input is a pandas dataframe.)

policytool/refparse/algo_evaluation/evaluate_match_references.py
lizgzil: changed to `actual_predicted_pairs` and newlines changed

policytool/refparse/algo_evaluation/evaluate_match_references.py
lizgzil: good idea

policytool/refparse/algo_evaluation/evaluation.md
lizgzil: good point, i've reworded this. although the default is still 10,000. I put it down to 1000 in investigate the threshold code

policytool/refparse/evaluate_algo.py
lizgzil: yeh I always run into relative path problems :/ 
I think I'll change them to relative in the new PR. It is neater and less text!

policytool/refparse/algo_evaluation/exploratory/investigate_match_thresholds.py
lizgzil: in investiagte_match_thresholds I have it as -1 so that it returns all the top matches for each reference, then I manually filter these matches using the different thresholds. It's quicker doing it this way than running the fuzzymatcher many times with different thresholds.

PR #168, opened 2019-07-01
Fix fuzzy match tests to work with refactored FuzzyMatcher

nsorros:
Fixes fuzzy matcher tests that were failing after the refactoring that was introduced in #156


hblanks:
Nice!

PR #169, opened 2019-07-01
Add a task in airflow to pull data from the epmc metadata bucket and store them in ES

SamDepardieu:
# Description
Add a task in airflow to pull data from the epmc metadata bucket and store them in ES.

## Type of change
- [x] :sparkles: New feature

# How Has This Been Tested?

`make test`
`make docker-test`
Test and full run on MacOS 10.14.5, Python3.7
Full run on a dev instances (m5.xlarge, 100GB HD, 16GBRam, 2VCPU)


Dockerfile.policytool
```diff
@@ -0,0 +1,40 @@
...
+COPY policytool/airflow/initdb.sh /airflow/
+
+# CodePipeline (but not CodeBuild) code checkouts lose execution bits
+# (https://forums.aws.amazon.com/thread.jspa?threadID=235452).
+# Fix this, link the dags into place, and set ownership.
+RUN chmod +x /airflow/initdb.sh && \
+    ln -s /src/policytool/airflow/dags /airflow/dags && \
+    chown -R www-data: /airflow
+
+USER www-data
```
hblanks: Hmm. I'm not sure I follow this. Could we keep the `base-image` target and update the existing `Dockerfile` as needed?

docker-compose.yml
```diff
@@ -79,26 +76,26 @@ services:
     environment:
       - discovery.type=single-node
       - bootstrap.memory_lock=true
-      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+      - "ES_JAVA_OPTS=-Xms1g -Xmx1g"
     ports:
       - 127.0.0.1:9200:9200
     volumes:
-      - esdata:/usr/share/elasticsearch/data
+      - ./policytool/elastic/elastic/db:/usr/share/elasticsearch/data
```
hblanks: This belongs in `build/` in the same way as how we keep the dev postgres DB there.

docker-compose.yml
```diff
@@ -157,8 +154,8 @@ services:
       - --reload
       - policytool.web.wsgi:application
     volumes:
-      - ./policytool/web:/src/policytool/policytool/web/
-      - ./policytool/scraper:/src/policytool/policytool/scraper/
+      - ./policytool/web:/src/policytool/web/
+      - ./policytool/scraper:/src/policytool/scraper/
```
hblanks: I'm not sure these path changes follow -- we should be volume'ing the entire repo into the container under `/src/policytool`, not just the python package.

policytool/airflow/tasks/fetch_epmc_metadata.py
```diff
@@ -0,0 +1,74 @@
...
+from policytool.elastic.import_epmc_metadata import (
+    import_into_elasticsearch,
+    clean_es,
+)
+
+
+class FetchEPMCMetadata(BaseOperator):
+    """ Bulk downloads all metadata XML from EPMC, writing results to
+    json.gz in S3.
+    """
```
hblanks: Docstrings are for a different task than what this is -- the fetch took place in the open research DAG.

Task name is also copied from that task. Maybe `IndexPublications` or `ESIndexPublications` or such for a name?

policytool/elastic/import_epmc_metadata.py
```diff
@@ -0,0 +1,182 @@
...
+        pub_list: A list containing both actions and data to be executed by
+                  Elasticsearch's bulk API
+    """
+    pub_list = []
+    for index, metadata in enumerate(yield_publications_metadata(s3_object)):
+        pub_list.append(metadata)
+        if max_publications and index + 1 >= max_publications:
+            yield pub_list
+            pub_list = []
+            break
```
hblanks: Probably cleaner to just yield and then `return`.

Makefile
```diff
@@ -59,10 +59,11 @@ base-image:
 		.
 
 .PHONY: docker-build
-docker-build: base-image $(WEB_BUILD_TARGETS)
+docker-build: base-image
```
hblanks: What happened to the web build targets?

docker-compose.yml
```diff
@@ -6,12 +6,13 @@ x-airflow-image: &airflow-image
...
 
-  # AIRFLOW_CONN_DIMENSIONS_DEFAULT: "https://${DIMENSIONS_USERNAME}:${DIMENSIONS_PASSWORD}@app.dimensions.ai/api/"
+
 
   # S3 prefix for all datalabs operators.
   # (in `core` because we must be in an existing section of the cfg)
-  AIRFLOW__CORE__DATALABS_S3_PREFIX: "s3://datalabs-dev/policytool-airflow" # No trailing /!
-  # AIRFLOW__CORE__DATALABS_ALTERYX_PREFIX: "s3://datalabs-staging/alteryx" # No trailing /!
+  AIRFLOW__CORE__POLICYTOOL_S3_PREFIX: "s3://datalabs-dev/policytool-airflow" # No trailing /!
+  AIRFLOW__CORE__DATALABS_S3_PREFIX: "s3://datalabs-dev/airflow" # No trailing /!
```
hblanks: These will probably conflict

docker-compose.yml
```diff
@@ -26,20 +27,16 @@ x-airflow_env: &airflow-env
...
   SENTRY_DSN: "${SENTRY_DSN}"
+
   DATABASE_URL: "postgresql://postgres:development@postgres:5432/wsf_scraping"
   RESOURCES_FILES: "/src/policytool/resources/"
 
 x-airflow-volumes: &airflow-volumes
   - ./policytool/airflow/airflow.cfg:/airflow/airflow.cfg
   - ./policytool/airflow/initdb.sh:/airflow/initdb.sh
-  - ./policytool:/src/policytool/policytool
+  - ./policytool:/src/policytool
```
hblanks: This still probably shouldn't be changed

docker-compose.yml
```diff
@@ -75,30 +72,29 @@ services:
       - ./build/airflow.redis:/data
 
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:7.0.0
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.2.0
     environment:
       - discovery.type=single-node
-      - bootstrap.memory_lock=true
-      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+      - "ES_JAVA_OPTS=-Xms1g -Xmx1g"
```
hblanks: I'd probably aim for 2GB

hblanks: good choice moving it up

policytool/elastic/import_epmc_metadata.py
```diff
@@ -0,0 +1,186 @@
...
+
+    Args:
+        es: a living connection to elacticsearch
+        s3_file: An open StreamingBody from s3
+        max_epmc_metadata: The maximum publication number to be inserted
+    """
+
+    with ThreadPool(THREADPOOL_SIZE) as pool:
+        if THREADPOOL_SIZE > 1:
+            pool_map = pool.map
```
hblanks: This should be `pool.imap()` for lower memory usage.

docker-compose.yml
SamDepardieu: Yeah this is in a later commit, forgot to push it

policytool/elastic/import_epmc_metadata.py
SamDepardieu: Interesting! It also makes the task non-blocking though, so it implies a bit of rewriting

SamDepardieu:
Yeah, the requierments update was exactly about that. I wanted to fix the vulnerabilities github found.
Everything (apart from Jinja?) has been fixed now, though. So I can remove this if you prefer!

hblanks:
Either way is OK there, just we for commit messages to explain changes like this.

policytool/elastic/import_epmc_metadata.py
```diff
@@ -109,7 +116,9 @@ def process_es_bulk(pub_list, es):
         request_timeout=3600,
     )
     if bool(bulk_response.get('errors')):
-        logger.info(bulk_response)
+        logger.error('failed on bulk indexing:\n%s',
+                     bulk_response)
+        raise IndexingErrorException()
```
hblanks: Thanks!

Note the `bool` is unnecessary.

hblanks: (Both none and empty list are false-y)

SamDepardieu: Here bulk_response['errors'] just returns the string "True" or "False", hence the conversion

SamDepardieu: Also my condition is clearly not working. Fixing this quick before merging

PR #198, opened 2019-07-30
Add a task to the pipeline to index all scraped fulltexts

SamDepardieu:
# Description

This PR aims to add a step to the pipeline to index all pdf fulltexts, parsed from policy documents, into an elasticsearch database.

To this purpose:
 - Add the import_fulltext_s3 script, usable out of airflow
 - Add a task in airflow using this script
 - Add the task to the policy dag

Also:
 - Fix a minor import path issue (`.pdf_parser.pdf_parse -> .pdf_parse`)
 - Fix the return of the epmc index method (Removes a /2, as the instruction and data are one element of the list, not two)
 - Add the `OPENRESEARCH_S3_PREFIX` variable again to docker compose, as it was needed by the index epmc task

## Type of change

- [x] :sparkles: New feature

# How Has This Been Tested?

`make test`
`make docker-test`

Local runs using Docker, docker-compose (@10Gb RAM and 2 VCPU)


policytool/airflow/dags/policy.py
```diff
@@ -94,7 +107,8 @@ def create_extract_pipeline(dag, organisation, spider_op_cls):
             dag, 'policy-extract', organisation, '.json.gz'),
         dag=dag)
 
-    spider >> parsePdfs >> extractRefs
+    es_index_fulltexts << parsePdfs
```
hblanks: May have better legibility if we run all arrows in the same order, please, with later steps below early ones. I.e.:

```python
es_index_publications >> ...
parsePdfs >> es_index_fulltexts
```

Also, I don't think the spider actually depends on loading publications into ES. In that sense, `create_extract_pipeline()` probably doesn`t need to take `es_index_publications` as an arg.

policytool/airflow/dags/policy.py
```diff
@@ -125,14 +139,13 @@ def create_dag(default_args, spider_op_cls):
         dag=dag
     )
     for organisation in ORGANISATIONS:
-        extract_task = create_extract_pipeline(
```
hblanks: May as well keep the assignment -- all these extract tasks are going to be made upstreams for the matcher(s).

policytool/airflow/tasks/es_index_fulltext.py
```diff
@@ -0,0 +1,84 @@
...
+
+from policytool.airflow.hook.wellcome_s3_hook import WellcomeS3Hook
+from airflow.models import BaseOperator
+from airflow.utils.decorators import apply_defaults
+from elasticsearch import Elasticsearch
+
+from policytool.elastic.import_fulltexts_s3 import (
+    import_into_elasticsearch,
+    clean_es,
+)
```
hblanks: Please group imports by standard lib, 3rd party, and internal, and sort. You may also want to just import twice from policytool.elastic, also -- the multiline import tuples tend to invite endless additional imports when pulling in the module wholesale is usually right. E.g.:

```python
from airflow.models import BaseOperator
from airflow.utils.decorators import apply_defaults
from elasticsearch import Elasticsearch

from policytool.airflow.hook.wellcome_s3_hook import WellcomeS3Hook
from policytool.elastic.import_fulltexts_s3 import import_into_elasticsearch
from policytool.elastic.import_fulltexts_s3 import clean_es
```

policytool/elastic/import_fulltexts_s3.py
```diff
@@ -0,0 +1,218 @@
...
+    Args:
+        es: a living connection to elacticsearch
+        bulk_query: a formatted bulk query to submit to Elasticsearch.
+    """
+    bulk_response = es.bulk(
+        body=''.join(pub_list),
+        refresh='wait_for',
+        request_timeout=3600,
+    )
+    if bulk_response.get('errors') == "True":
```
hblanks: I may be missing something, but the [ES bulk API docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) suggests that it may be better to check thus:

```python
if not bulk_response['errors']:
```

since it appears to be a boolean.

policytool/elastic/import_fulltexts_s3.py
```diff
@@ -0,0 +1,218 @@
...
+        raise IndexingErrorException()
+    return len(pub_list)
+
+
+def clean_es(es):
+    """ Empty the elasticsearch database.
+
+    Args:
+        es: a living connection to elasticsearch
+
```
hblanks: Extra newline

policytool/elastic/import_fulltexts_s3.py
```diff
@@ -0,0 +1,218 @@
...
+    Args:
+        es: a living connection to elacticsearch
+        bulk_query: a formatted bulk query to submit to Elasticsearch.
+    """
+    bulk_response = es.bulk(
+        body=''.join(pub_list),
+        refresh='wait_for',
+        request_timeout=3600,
+    )
```
hblanks: Out of curiousity, did we consider using [elasticsearch.helpers.bulk()](https://elasticsearch-py.readthedocs.io/en/master/helpers.html#elasticsearch.helpers.bulk) or its ilk, such as `parallel_bulk()`? These seems to do the chunking for us, plus threads, plus raising at the end of all indexing if any errors occurred.

policytool/elastic/import_fulltexts_s3.py
```diff
@@ -0,0 +1,218 @@
...
+        'hash': line['file_hash'],
+        'text': line['text'],
+        'organisation': org,
+    })
+
+    data = body + '\n'
+    return '\n'.join([action, data])
+
+
+def yield_publications(s3_object, org):
```
hblanks: Please consider making these functions take file objects instead of S3 objects, so that they can be called using anything supporting Python's file methods.

policytool/airflow/tasks/es_index_fulltext.py
```diff
@@ -0,0 +1,84 @@
...
+                self.src_s3_key,
+            )
+        else:
+            self.log.info(
+                'Getting %s pubs from %s',
+                'all',
+                self.src_s3_key,
+            )
+
+        s3_file = s3.get_key(self.src_s3_key)
```
hblanks: Practice in all our tasks has been to download to a temporary file so that if the consumer is too slow, the connection to S3 will not time out on us. Can we do that here as well, please?

policytool/elastic/import_fulltexts_s3.py
SamDepardieu: It is indeed a boolean! Surprising because I made this change because after changing it to a boolean check, it passed as true everytime... 
Anyway, I retried it and it's working. It was probably just a bad volume synchronisation between local and docker compose environments 👍

policytool/elastic/import_fulltexts_s3.py
SamDepardieu: I saw the helpers when I came at the end of the previous elasticsearch indexing task. I'm not too sure yet how big of a change it is, but I can see clear benefits. Do you want me to take some time to use it instead of our current code?

hblanks: Let's merge. Can you please add a trello card to do this, though, and put it in the next sprint column?

SamDepardieu:
Card created in trello, merging this

PR #209, opened 2019-08-21
Integrate reference splitter

nsorros:
To do
* ~~install reference_splitter in Docker~~
*  ~~fix tests~~
* ~~fix evaluation~~
* ~~specify the URL & version of this wheel in only a single file~~

## Evaluation update:
Evaluation of splitter updates to: 
```
'Score': 0.757,
'Number of reference sections in sample': 70,
'Proportion below threshold of 40':0.757,
'Median difference': 20.0,
'Metrics grouped by source': 
Source                                     msf   nice  unicef  who_iris
Number of reference sections in sample   9.000  25.00    14.0    22.000
Median difference                       33.300  16.70    37.1    13.900
Proportion below threshold of 40         0.667   0.92     0.5     0.773
```
which is more or less what we'd expect from 'Test Metric of Current Model on Policy Test Data' [here](https://github.com/wellcometrust/datalabs/blob/master/modelling/policy/reference-splitter/docs/accuracy.md#test-metric-of-current-model-on-policy-test-data) (note that this analysis was done when the sample size was 53, now it is 70)

## Tests
```
----------------------------------------------------------------------
Ran 27 tests in 0.027s

OK
```


nsorros:
@lizgzil I will fix the first item in the todo list and leave the rest to you. I would hapilly fix the tests but the new splitter is behaving a bit weird for the use cases in the test so maybe it is a good time to ask ourselves what do we want/expect from it?

For example, when passed an empty string it fails due to this line `if references_section[0] != "\n":` and also when it does not find a begging of a reference it returns empty list so when passed "This" it returns []. The good news is that the tests helped us identify these cases easily.

lizgzil:
> @lizgzil I will fix the first item in the todo list and leave the rest to you. I would hapilly fix the tests but the new splitter is behaving a bit weird for the use cases in the test so maybe it is a good time to ask ourselves what do we want/expect from it?
> 
> For example, when passed an empty string it fails due to this line `if references_section[0] != "\n":` and also when it does not find a begging of a reference it returns empty list so when passed "This" it returns []. The good news is that the tests helped us identify these cases easily.

@nsorros do you mean 'evaluations' instead of 'tests', or are the unit tests the things that are failing? Yep, I will pick this up when I'm back up :)

nsorros:
> > @lizgzil I will fix the first item in the todo list and leave the rest to you. I would hapilly fix the tests but the new splitter is behaving a bit weird for the use cases in the test so maybe it is a good time to ask ourselves what do we want/expect from it?
> > For example, when passed an empty string it fails due to this line `if references_section[0] != "\n":` and also when it does not find a begging of a reference it returns empty list so when passed "This" it returns []. The good news is that the tests helped us identify these cases easily.
> 
> @nsorros do you mean 'evaluations' instead of 'tests', or are the unit tests the things that are failing? Yep, I will pick this up when I'm back up :)

I mean both

Makefile
```diff
@@ -81,6 +81,7 @@ $(VIRTUALENV)/.installed: requirements.txt test_requirements.txt
 	@mkdir -p $(VIRTUALENV)
 	virtualenv --python python3 $(VIRTUALENV)
 	AIRFLOW_GPL_UNIDECODE=yes $(VIRTUALENV)/bin/pip3 install -r requirements.txt
+	$(VIRTUALENV)/bin/pip3 install https://datalabs-public.s3.eu-west-2.amazonaws.com/references_splitter/reference_splitter-2019.8.0-py3-none-any.whl
```
hblanks: Looks good. @lizgzil  - please remind me before merge to see if I can't update this PR so that we specify the URL & version of this wheel in only a single file, which then is somehow used to make both the docker image (change above) and the localdev virtualenv here in the Makefile.

hblanks: (in the trade we call this principle DRY - Don't Repeat Yourself)

nsorros: i am wondering whether you are thinking of a common bash script for this?

hblanks: More likely that we will specify the wheel URL as a variable in the `Makefile`, then use that both here in this target, as well as passing into the `docker build` command as an environment variable.

Well, more likely yet is to specify the wheel URL in its own file, which both the Makefile and Dockerfile then depend on.

So, close to what you say, but we'd want to just store the data in this file, not how it's used (as it would be if it were a bash script..though sometimes that's the right choice, too)

Dockerfile.base
```diff
@@ -22,4 +22,5 @@ RUN apt-get update -yqq && \
     locale-gen C.UTF-8 && \
     pip install -U pip && \
     AIRFLOW_GPL_UNIDECODE=yes python3 -m pip install -r requirements.txt && \
+    pip install https://datalabs-public.s3.eu-west-2.amazonaws.com/references_splitter/reference_splitter-2019.8.0-py3-none-any.whl && \
```
lizgzil: @nsorros am I right in thinking we should delete the references_splitting folder from here now: https://s3.console.aws.amazon.com/s3/buckets/datalabs-public/?region=us-east-2&tab=overview

Makefile
lizgzil: To run this locally I had to run `pip install https://datalabs-public.s3.eu-west-2.amazonaws.com/references_splitter/reference_splitter-2019.8.0-py3-none-any.whl` (otherwise I get `ModuleNotFoundError: No module named 'reference_splitter'`). So it doesn't seem like it's been added to the virtualenv packages (unless I'm doing something wrong!), any ideas?

lizgzil:
@ivyleavedtoadflax @nsorros @aCampello how do we want the reference splitter to work? I.e. what should we always be outputting?
This is what we had in the regex tests:

**Case 1: no references section text** 
```
split_section("")
```
should return `[]`

**Case 2: just one line of text**
```
split_section("This is one line")
```
should return `["This is one line"]`
**Case 3: just special characters, no text**
```
split_section("\n\n\n")
```
should return `[]`

**Case 4: multiple references text**
```
split_section("1. One reference\n2. Two references\n3. Three references\n")
```
should return `['One reference', 'Two reference', 'Three reference']`

### My ideas:
I think case 1 and 3 are still as they should be. But with the new IOBE spitter case 2 now gives `[]` and case 4 gives `['1. One reference', '\n2. Two references', '\n3. Three references\n']`.

In the new ML ways of splitting references it will require the example text to actually look like references, so I propose we scrap cases 2 and 4 and have:
```
len(split_section("Smith et al. 2019. This is a title. Journal of journals. 1-2"))
```
should return 1, and

```
len(split_section("References \n1. Smith et al. 2019. This is a title. Journal of journals. 1-2. \n2. Smith et al. 2019. This is a title. Journal of journals. 1-2. \n3. Smith et al. 2019. This is a title. Journal of journals. 1-2."))
```
should return 3.

### Problem
It's hard to not overfit these examples so that they work for the IOBE line splitter, but there are some weird things, e.g.
`len(split_section("Smith et al. 2019. This is a title. Journal of journals. 1-2"))` returns 1, but `len(split_section("Smith et al. 2019. This is a title. Journal of journals. 1-2."))` (i.e. full stop at the end) returns 0, and
`len(split_section("Smith et al. 2019. This is a title. Journal of journals. 1-2. \n"))` returns 1.

What do you think ? Can you see any problems or any other tests to include?

ivyleavedtoadflax:
I agree with all your suggestions here @lizgzil . *Case 2* should only return something when it finds a real reference, and since I'm assuming we will use the new model by passing an entire document text to it, *Case 4* should no longer be relevant.

I think each system is going to have its own idiosyncrasies, so I expect that this is going to be a work in progress that will need updating after each future update.

hblanks:
Nice - thanks for working on the reference splitter wheel URL with me!

hblanks:
@lizgzil  - LMK when you're good to merge!

PR #211, opened 2019-08-29
Fix a typo in imports

SamDepardieu:
All in title


hblanks:
May be good next sprint to make a test that just imports the DAG...though to do that, you have to mock out a bunch of airflow crud.

PR #215, opened 2019-08-30
Write a dict instead of null for unparsed pdfs

SamDepardieu:
# Description

The pdf parser wrote null to the output file when it couldn't parse a
pdf. This caused the next task to sometime be unable to read the json
file.

(See #214 )

## Type of change

### DIFF
--- a/policytool/pdf_parser/main.py
+++ b/policytool/pdf_parser/main.py
@@ -95,7 +95,14 @@ def parse_pdf(pdf, words, titles, context, pdf_hash):
             logger.warning('The pdf couldn\'t be parsed {pdf_hash}'.format(
                 pdf_hash=pdf_hash,
             ))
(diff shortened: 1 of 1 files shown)


policytool/pdf_parser/main.py
```diff
@@ -95,7 +95,14 @@ def parse_pdf(pdf, words, titles, context, pdf_hash):
             logger.warning('The pdf couldn\'t be parsed {pdf_hash}'.format(
                 pdf_hash=pdf_hash,
             ))
-            return None
+
+            # We still have to return something for the json to be complete.
+            return {
+                'file_hash': pdf_hash,
```
hblanks: Hmm. May be best to just not include these files at all. But, can see why we'd want them, too. So, let's go ahead with it.

PR #218, opened 2019-09-03
Add some fields to the scraper's manifest

SamDepardieu:
# Description
Add's start-time, end-time and organisation fields to the manifest

Also ignore those fields in the pdf parser task.

## Type of change
- [x] :sparkles: New feature

# How Has This Been Tested?
Local tests
`make test`
`make docker-test`

### DIFF
--- a/policytool/pdf_parser/main.py
+++ b/policytool/pdf_parser/main.py
@@ -21,6 +21,13 @@ logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
 
+MANIFEST_IGNORE_FIELDS = [
+    'start-time',
+    'stop-time',
+    'organisation',
+]
+
+
 def default_resources_dir():
     """ Returns path to resources/ within our repo. """
     return os.path.join(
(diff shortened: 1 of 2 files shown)


policytool/pdf_parser/main.py
```diff
@@ -231,21 +238,25 @@ def parse_all_pdf(input_url, output_url, organisation, context,
...
-                item = parse_pdf(
-                    tf.name,
-                    words,
-                    titles,
-                    context,
-                    item,
-                )
-            parsed_items.append(item)
+        if directory in MANIFEST_IGNORE_FIELDS:
+            continue
```
hblanks: Could we nest the manifest instead, please? I.e.:

```python
manifest = {
    'metadata': {...}',
    'content': {
        'aa': [],
    }
}
```

or such?

SamDepardieu:
@hblanks Updated and tested - clean and working 👍

policytool/scraper/wsf_scraping/file_system.py
```diff
@@ -118,15 +120,27 @@ def get_manifest(self):
 
     def update_manifest(self, data_file):
         current_manifest = self.get_manifest()
+
+        metadata = {}
+        metadata['organisation'] = self.organisation
+        metadata['start-time'] = \
+            self.start_time.strftime("%Y-%m-%d, %H:%M:%S")
+        metadata['stop-time'] = \
+            datetime.datetime.now().strftime("%Y-%m-%d, %H:%M:%S")
```
hblanks: Can you please update this to ISO8601? Easiest way would be `self.start_time.isoformat()`, `datetime.datetime.now().isoformat()`?

hblanks:
One more note - do we need to update anything downstream as well, or does the filesystem take care of this? (Seems likely we might need to..)

SamDepardieu:
Changed to iso as requested. What do you mean by downstream? The only consumer of this manifest at the moment is the pdf parser task, and this is already handled in this PR

PR #228, opened 2019-09-09
pdf_parser: skip pdf2html outputs > 64MB

hblanks:
# Description

Some PDF files have caused our airflow worker to OOM. Remediate this
by skipping these files (which, on inspection of one at least, didn't
contain any useable data):

* Skip PDF files where pdf2html output is > 64MB
* Update parse_pdf_document() so it returns parse errors
* Include PDF parse errors in JSON output, so we can track how many
  parses fail and of what kind
* ~~Use LXML instead of BeautifulSoup for parsing HTML (faster, may be
  more memory efficient)~~

Issue: https://github.com/wellcometrust/policytool/issues/227

## Type of change

# How Has This Been Tested?

```
./docker_exec.sh airflow test policy ParsePdf.who_iris 2019-09-01
```

using a `docker-compose.yml` that pointed to `s3://datalabs-staging/`, and where the manifest was updated to only include files with hash prefix of `69`.

`make docker-test`

# Checklist:

### DIFF
--- a/policytool/pdf_parser/main.py
+++ b/policytool/pdf_parser/main.py
@@ -91,12 +91,14 @@ def parse_pdf(pdf, words, titles, context, pdf_hash):
         item: A dict containing the pdf text, sections and keywords.
     """
     # Convert PDF content to text format
-    pdf_file, pdf_text = parse_pdf_document(pdf)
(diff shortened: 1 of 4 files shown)


hblanks:
(The LXML change actually breaks the unit tests...so what we have there isn't exactly the same. Defer that part for later)

PR #240, opened 2019-09-18
Use single index for policy docs; add cli

hblanks:
# Description

Two small changes:

1. Use the same index name for all policy documents we index.
1. Add a minimal CLI for obtaining an ElasticSearch client, via `python -i -m policytool.elastic.common`

## Type of change

(feature?)

# How Has This Been Tested?

1. Running test & main DAG on EC2 instance.
1. `make docker-test`

# Checklist:

### DIFF
--- a/policytool/elastic/common.py
+++ b/policytool/elastic/common.py
@@ -204,3 +204,11 @@ def insert_from_argv(description, clean_es, insert_file):
     else:
         with open(args.input, 'rb') as f:
             return insert_file(f, es, args.max_items)
+
+
+if __name__ == '__main__':
+    parser = create_argument_parser(
+        'Instantiates ES client for use with python -i')
+    args = parser.parse_args()
+
+    es = es_from_args(args)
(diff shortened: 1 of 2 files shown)


aCampello:
Looks good to me. Seems to be generating one single index called `policy-docs` instead of one for each document.

policytool/elastic/common.py
```diff
@@ -204,3 +204,11 @@ def insert_from_argv(description, clean_es, insert_file):
...
         with open(args.input, 'rb') as f:
             return insert_file(f, es, args.max_items)
+
+
+if __name__ == '__main__':
+    parser = create_argument_parser(
+        'Instantiates ES client for use with python -i')
+    args = parser.parse_args()
+
+    es = es_from_args(args)
```
aCampello: Seems that you'll only to be able to set `port` and `host` with this CLI - but probably that's enough for your purposes.

hblanks: correct! This CLI's really just a convenience for dev & ad-hoc ops against Elasticsearch.

PR #248, opened 2019-10-03
Add urls back to the manifest file

### LINKED ISSUE #244: Add urls to scrape data
As far as I am aware from looking at the things saved in reach-airflow, at the moment the only things saved about the policy documents are ['file_hash', 'sections', 'keywords', 'text'].

I'd say it is very important to save the urls either here or in a metadata JSON (in which year and title would also be useful), perhaps even for the alpha release. Otherwise the policy document data has no real-world link anymore.

SamDepardieu:
# Description
The lack of url attribute is blocking right now (#244 ) so this PR adds a 'data' section to the manifest file as a temporary fix.
This is non-optimal and need changes, as it duplicates the file_hash as dict key. On the other hand, it doesn't affect anything else on the DAG so everything else can run normally.

Fix #244 

## Type of change

- [x] :sparkles: New feature

# How Has This Been Tested?
`make docker-test`
Local runs on Docker with 2CPU, 10.5GiB:
 - `./docker_exec.sh airflow test policy-test Spider.msf 2019-10-02`
 - Verification of the produced json file

### DIFF
--- a/policytool/scraper/wsf_scraping/file_system.py
+++ b/policytool/scraper/wsf_scraping/file_system.py
@@ -132,6 +132,7 @@ class S3FileSystem(FileSystem):
 
         # If the manifest has no content yet, let's create it
         content = current_manifest.get('content', {})
+        data = current_manifest.get('data', {})
         for row in data_file:
             item = json.loads(row)
             hash_list = content.get(item['hash'][:2], None)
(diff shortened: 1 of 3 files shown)


policytool/scraper/wsf_scraping/file_system.py
```diff
@@ -141,6 +142,9 @@ def update_manifest(self, data_file):
             else:
                 content[item['hash'][:2]] = [item['hash']]
 
+            if not data.get(item['hash']):
+                data[item['hash']] = {'url': item['url']}
```
lizgzil: should something be in place in case `item` doesn't have 'url'? Or are you definitely sure it will have it?

SamDepardieu: I'm definitely sure every item have a url at this point. If it doesn't, it'd be good to get an error here anyway, because that's really unexpected

Issue #183, opened 2019-07-18
Investigate why there are duplicate ref matches found in the same policy document


lizgzil:
Duplicate reference found even though no duplicate found in policy document.

A citation for "Trends in adult body-mass index in 200  countries from 1975 to 2014: a pooled analysis of  1698 population-based measurement studies with   1377-1396" only occurred once in a policy document (http://www.fao.org/3/I9553EN/i9553en.pdf), but the Reach output said it came up twice.

the tool also said it came up twice in
https://www.unicef.org/publications/files/State_of_Food_Security_and_Nutrition_in_the_World_2017.pdf
but it only occurs once.

"Disability-adjusted life years (DALYs) for 291 diseases and injuries in 21 regions, 1990‚Äì2010: a systematic analysis for the Global Burden of Disease Study 2010" is said to come up twice in 
https://apps.who.int/iris/bitstream/handle/10665/144099/Early%20Child%20Development%20in%20the%20European%20Region%20needs%2c%20trends%20and%20policy%20development.pdf?sequence=1&isAllowed=y
but it actually only comes up once.

## November update:

Just to note that this is still an issue. I ran the exact matcher using the policy dag and got 7 or 9 duplicate matches for each match.
![Screenshot 2019-11-08 at 14 53 55](https://user-images.githubusercontent.com/15956065/68485682-a309b280-0237-11ea-9525-6cc7f11e38ea.png)

Looking into this issue more I printed the results of the elastic search when there was a match found, and I see that there are indeed the same policy doc hashes but with different ES indices.

e.g. here is an example of the matches found when searching for one EPMC publication title in the text of policy documents:
```
title = publication['title']

        # Exclude short titles that create noisy matches
        if len(title) < self.title_length_threshold:
            return

        body = {
            "query": {
                "match_phrase": {
                    "doc.text": title
                }
            },
            "_source": ["doc.hash"],

        }
        res = self.es.search(
            index=self.es_full_text_index,
            body=body,
            size=1000 # if there are more than 1000, need to paginate
        )
```
```
'hits': 
[{'_index': 'policy-test-docs', '_type': '_doc', '_id': 'aXpUQW4BiUIv3WUbqiOh', '_score': 19.55179, '_source': {'doc': {'hash': '42db3dd8b3156d81c60378bf0e52350d'}}},
[...]

PR #252, opened 2019-10-07
Make exact matcher look in full text

### LINKED ISSUE #202: The exact matcher is only used in reference section texts
At the moment we have
```
exact_matcher = ExactMatcher(sectioned_documents, 
     settings.MATCH_TITLE_LENGTH_THRESHOLD)
```
in `refparse`, but actually for the exact matching _I think_ we should
1. search in **all** the pdf text
2. search regardless of whether the pdf has a references section or not.

lizgzil:
Fix #202 

# Description

This PR adapts the code to look for exact matches in all the policy text, rather than just in the references sections.

- Rename exact matcher variables to not have anything to do with sections
- Adds an optional argument in `get_scraping_results` in file_manager.py which allows you to input which columns to include in the scraper output dictionary. This will mean we can include the text column needed to exact match on all the text. I did this rather than just updating the global variable `SCRAPING_COLUMNS` with 'text' since for the fuzzy match it is unneeded to store the large text data in memory.
- I create a new function `transform_scraper_text_file` in `refparse.py` to specifically transform the scraped full text for the exact match. This function is v simliar to `transform_scraper_file`, but I chose to create another function rather than using an if statement in the same one - perhaps this would be better though?
- Added the exact matcher task to the policy dag (editing `exact_match_refs_operator.py` and `policy.py`).
- Adds 8 new exact matcher tests
- Changes to `test_pdf_objects` to allow for these tests to be run (or skipped) on a mac (this should be fixed properly at some point, see issue #273 )

## Type of change

## Running time increases a lot!

A ran locally for the 70 documents in the latest MSF scrape, using:
```
python -m policytool.refparse.refparse \
    --scraper-file "s3://datalabs-dev/reach-airflow/output/policy/parsed-pdfs/msf/parsed-pdfs-msf.json.gz" \
    --references-file "s3://datalabs-data/wellcome_publications/uber_api_publications.csv" \
    --model-file "s3://datalabs-data/reference_parser_models/reference_parser_pipeline.pkl" \
```
this took **504 seconds** and finds 10 doc-publication matches.

Without these changes this took **8 seconds** and found 2 doc-publication matches (reassuring also found in the 10 full text search matches).

- I tried flashtext to make it quicker, but it didn't work on looking for whole sentences in large amounts of text - it works for 1 or 2 words in text.
- I tried Whoosh to make this quicker, but it took 1774 seconds and found 7538 matches - obviously there is probably a bug somewhere, but I didn't investigate any further.
- I tried Spacy's PhraseMatcher (`pip install spacy` and `python -m spacy download en_core_web_sm`) but aborted trying this out, since in airflow we dont need it, so it doesnt matter if refparse ran locally is slow

I've created an issue #257, if someone wants to pick it up this performance issue in the future.

# How Has This Been Tested?

```
>>> make docker-test
74 passed, 4 warnings in 10.56s
>>> make test
 72 passed, 2 skipped, 5 warnings in 7.34s
>>> cd reach/refparse/
>>> python -m unittest
Ran 35 tests in 0.112s

OK
```
I ran:
```
source build/virtualenv/bin/activate
eval $(./export_env.py)
docker-compose up -d
./docker_exec.sh airflow test policy-test ExactMatchRefs.msf 2019-11-8
```
it took about 3 hours to run and processed 5377194 publications and found 8132 matches. 
I noticed my results have duplicates in, fixing this might help speed things up if we delete duplicate policy docs before the exact matcher runs. I added some info about this duplication to issue #183


SamDepardieu:
> this took 504 seconds and finds 10 doc-publication matches.

> Without these changes this took 8 seconds and found 2 doc-publication matches (reassuring also found in the 10 full text search matches).

I don't see anything obvious, but this is slightly worrying in my opinion. 504 seconds for 70 documents is a lot, especially considering that some of our organisations (e.g. parliament) have around 60k documents.

We may want to profile this code to find out what's taking so long and to see if we can come with a solution for it to run faster before merging this PR.

I'd be happy to help you with that if needed, though

lizgzil:
@SamDepardieu 
> I'd be happy to help you with that if needed, though
That'd be great! Yes, it's quite a huge increase in time :/

policytool/refparse/refparse.py
```diff
@@ -62,6 +62,22 @@ def transform_scraper_file(scraper_data):
...
+def transform_scraper_text_file(scraper_data, text_column="text"):
+    """Takes a pandas dataframe. Yields back individual
+    SectionedDocument tuples.
+    """
+    for _, document in scraper_data.iterrows():
+        if document[text_column]:
+            try:
+                text = document[text_column]
+            except KeyError:
+                return
```
nsorros: this will stop the yielding if any of the rows does not have a text column

policytool/refparse/refparse.py
```diff
@@ -62,6 +62,22 @@ def transform_scraper_file(scraper_data):
...
+            yield SectionedDocument(
+                text,
+                None, # TODO: use document['uri'] after fixing scrape
```
nsorros: you can add it now by document.get('uri', None)

nsorros:
Also note that this change does not affect the Reach tool as you would need to change the airflow task exact match. You need to search in the full text index not only in reference section index. You should add this before we merge.

ivyleavedtoadflax:
There is a package I used in a previous role which is significantly faster than regex, and could be used in `ExactMatcher`. I'll see if I can dig it out!

ivyleavedtoadflax:
Found it We had good results with this. 

https://www.analyticsvidhya.com/blog/2017/11/flashtext-a-library-faster-than-regular-expressions/

https://github.com/vi3k6i5/flashtext

policytool/refparse/refparse.py
```diff
@@ -43,14 +43,14 @@ def check_publications_file(publications, publications_file):
...
     SectionedDocument tuples.
     """
     for _, document in scraper_data.iterrows():
-        if document["sections"]:
+        if document[section_column]:
             try:
-                sections = document['sections']
+                sections = document[section_column]
             except KeyError:
                 return
```
lizgzil: @hblanks could you check that these lines are performing as you think they should/are parts of it redundant? My first thought is that `if document[section_column]:` should make the `try` redundant (i.e. if it passes the `if` then it should also pass the `try` in this case)? And as @nsorros says below, if there is no 'sections' part of the dict for a document then this will stop all the rest of the documents being processed.

policytool/refparse/refparse.py
lizgzil: sam added this as 'url' in [his latest PR](https://github.com/wellcometrust/policytool/pull/248/files) so using that instead of 'uri'

lizgzil:
@SamDepardieu or @hblanks please can you check my commit changing the airflow stuff if you have time? Tests passed, but I'm not confident the arguments of `exact_match_refs_operator.ExactMatchRefsOperator` are correct.

lizgzil:
also (@ivyleavedtoadflax @SamDepardieu ) I've given up on making the local running of refparse quicker in this PR since it won't effect the product performance anyway. I'll create an issue, if someone wants to pick it up in the future.

policytool/airflow/tasks/exact_match_refs_operator.py
```diff
@@ -105,6 +104,7 @@ def execute(self, context):
     
         exact_matcher = ElasticsearchExactMatcher(
             self.es,
+            self.es_index,
```
nsorros: es_index is not a good name, all indices in elastic search are es_index. i suggest full_text_index or es_full_text index

nsorros:
> @SamDepardieu or @hblanks please can you check my commit changing the airflow stuff if you have time? Tests passed, but I'm not confident the arguments of `exact_match_refs_operator.ExactMatchRefsOperator` are correct.

Did you run the DAG? Did the exact matcher found matches?

Test pass as I think there are no tests testing the exact matcher.

SamDepardieu:
> > @SamDepardieu or @hblanks please can you check my commit changing the airflow stuff if you have time? Tests passed, but I'm not confident the arguments of `exact_match_refs_operator.ExactMatchRefsOperator` are correct.
> 
> Did you run the DAG? Did the exact matcher found matches?
> 
> Test pass as I think there are no tests testing the exact matcher.

The best way to test the dag is to run the `policy-test` DAG. If you're able to `docker-compose up -d` Reach, you can access Airflow's interface on `localhost:8080` and then just click the little "play" button on the right of the `policy-test` dag

hblanks:
@lizgzil  - yes, the place for debugging dags is unfortunately in the airflow web (and maybe scheduler logs), using `docker-compose logs airflow-web`, sometimes with options. It's yucky.

Your branch also could stand to be rebased on top of the reach rename. Would you like it if I did this and after we debugged the DAG together?

lizgzil:
> Your branch also could stand to be rebased on top of the reach rename. Would you like it if I did this and after we debugged the DAG together?

@hblanks Yes please, thatd be great!

lizgzil:
Thanks to @SamDepardieu and @hblanks I've fixed my issues with Airflow.

I ran `./docker_exec.sh airflow test policy-test ExactMatchRefs.msf 2019-11-8` and got 8132 matches in 3 hours. As noted in the description this takes longer than really necessary since there are duplicates in the ES EPMC Metadata index. I'm currently running the whole policy-test dag, but I anticipate it'll take a long time! (note msf is the smallest data and it still took 3hours).

It'd be great to get this merged so we could see if the policy-test dag succeeds not using my computer, so a review of the code / suggestions on how to speed up performance / help fixing issue #183 would be very much appreciated :)

hblanks:
@lizgzil  - thanks for keeping this going. A couple thoughts:

1. Because we use the test DAG to quickly verify the pipeline, the test DAG has to complete within a fairly small amount of time.
1. It doesn't sound like we can get the test DAG running quickly right away with the exact matcher in place.
1. But, we do want to have the exact matcher available in the codebase, and maybe even able to run in a deployment environment.

So, my two cents would be that we create a separate dag, `policy-test-exact-match`, which we can use to try speeding up things as best we can. The first place is probably batching search queries to ElasticSearch, if that can be done -- or else running them in a thread pool.

hblanks:
OK! This branch is rebased to a new branch, `fix-exact-matcher-rebase`. After adding a small commit, the exact matcher test DAG Liz added runs in short order. So, I suggest we:

1. Close this PR.
1. Create a PR for the new branch.
1. Maybe update `dag.py` so that it the main DAG doesn't run the exact matcher, but so that the test DAG does?

lizgzil:
Closing this and opening a rebased version https://github.com/wellcometrust/reach/pull/285

PR #262, opened 2019-10-25
Create the web search results page

SamDepardieu:
# Description

Implements the new search results API page, as well as fixing a few other design issues and refactoring the template rendering methods:

 - Add new routes for a templates search results page
 - Change search form action route
 - Changes the template class and architecture
   - The template class can now be used as a base class in other api calls
   - changes the template architecture to reflect the search/results routes of the api
 - Use Spectre base classes to define btn-xl and input-xl.
 - Changes and add to the design
   - Add a result page css file
   - Move some of the search page css to the global style design
   - Define some sizes for the new xl sized fields
   - Define a table-bordered class (similar to bootstrap's)
   - a few other minor tweaks
 - Fix the gulpfile 

## Type of change

- [x] :sparkles: New feature

# How Has This Been Tested?

`make docker-test`
Local runs on Docker with 2CPU, 10.5GiB memory

### DIFF
--- a/policytool/web/api.py
+++ b/policytool/web/api.py
@@ -11,12 +11,13 @@ import os.path
 from urllib.parse import urlparse
 
 import falcon
-import jinja2
 from elasticsearch import Elasticsearch
 
+from .templates import TemplateResource
 from . import search
 
 TEMPLATE_ROOT = os.path.join(os.path.dirname(__file__), 'templates')
+
 
 def configure_logger(logger):
     """ Configures our logger w/same settings & handler as our webserver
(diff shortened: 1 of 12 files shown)


policytool/web/templates.py
```diff
@@ -0,0 +1,74 @@
...
+    if path.endswith('.html'):
+        return (
+            path,
+            os.path.join(path[:-5], 'index.html'),
+        )
+
+    return (
+        path + '.html',
+        os.path.join(path, 'index.html'),
+    )
```
hblanks: 1. When you can, best to prefer singular names over plural. So, `template.py`.
1. Don't forget to move existing unit tests into place as well :-)

policytool/web/templates/index.html
```diff
@@ -25,7 +25,7 @@ <h3>Find research publications cited by Policy Docs</h3>
         <h3>Search through Policy Documents</h3>
         <p>We have pulled in the PDFs of UNICEF, MSF, NICE, WHO and the UK Parliament. Our tool allows you to search across these documents.</p>
         <div class="buttons">
-          <a class="btn btn-primary hero-button" href="/search">
+          <a class="btn btn-primary hero-button" href="/search/content">
```
hblanks: Probably should be `policy-docs` for the time being.

policytool/web/api.py
```diff
@@ -135,12 +67,21 @@ def create_api(conf):
...
         TemplateResource(TEMPLATE_ROOT, get_context(os.environ))
     )
+    api.add_route(
+        '/search/content',
+        TemplateResource(TEMPLATE_ROOT, get_context(os.environ))
+    )
+    api.add_route(
+        '/results/content',
+        search.FulltextPage(TEMPLATE_ROOT, es, conf.es_explain,
+                            get_context(os.environ))
```
hblanks: It may be a bit more common practice to only have:

```python
api.add_route(
    '/search/policy-docs',
    search.FulltextPage(TEMPLATE_ROOT, es, conf.es_explain, ...)
)
```

where if no params are present, you get a page akin to what's currently `/search/content`, and where when search params are present, you get the (nearly) same page but with search results below it (what's currently `/results/content`).

policytool/web/templates.py
SamDepardieu: Yup, true 👍

SamDepardieu: I was just updating them at the moment, because I re-ran them to be sure and some aren't passing

policytool/web/api.py
SamDepardieu: So changing `/search/content` to `/search/policy-docs` and same for `/results/content` => `/results/policy-docs` ?

hblanks: Not quite. I'm suggesting that there be only two URLs:

1. `/search/policy-docs`. For initial search page and search results after search query is given for policy docs.
1. `/search/citations`. For initial search page and search results after search query is given for citations.

hblanks: (so, no `/results/.*`)

SamDepardieu: Oh okay! Yes, agreed, that would be better. I'll work on that tomorrow then.

PR #263, opened 2019-10-25
Fix Elasticsearch indices

SamDepardieu:
# Description

🚨 This will change the name of the indices used in Elasticsearch 🚨 

Fix a few incoherence on ES indices:

 - No more indices called "policy-policy-X" or "policy-policy-test-X", just "policy-X" or "policy-test-X"

 - Fixed an issue where we only kept the last organisation results (one task per organisation, but full index clean every task)

## Type of change

- [x] :bug: Bug fix

# How Has This Been Tested?

`make docker-test`

Local test and full runs on Docker with 2CPU, 10.5GiB memory

### DIFF
--- a/policytool/airflow/dags/policy.py
+++ b/policytool/airflow/dags/policy.py
@@ -36,12 +36,15 @@ ItemLimits = namedtuple('ItemLimits', ('spiders', 'index'))
 # Configuration & paths
 #
 
+
 def verify_s3_prefix():
     reach_s3_prefix = conf.get("core", "reach_s3_prefix")
     assert reach_s3_prefix.startswith('s3://')
     assert not reach_s3_prefix.endswith('/')
 
+
 verify_s3_prefix()
+
 
 def get_es_hosts():
     result = conf.get("core", "elasticsearch_hosts")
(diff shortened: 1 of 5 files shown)


policytool/airflow/dags/policy.py
```diff
@@ -109,7 +112,7 @@ def create_org_pipeline(dag, organisation, item_limits, spider_years):
         organisation=organisation,
         es_host='elasticsearch',
         item_limits=item_limits.index,
-        es_index='-'.join([dag.dag_id, 'policy-docs']),
+        es_index='-'.join([dag.dag_id, 'docs']),
```
hblanks: Are we renaming elsewhere as well from "policy documents" to "docs"? If not, may be good to keep this.

SamDepardieu: No, it only change the indices from `[policy | policy-test]-policy-docs` to `[policy | policy-test]-docs`. It never was `[policy | policy-test]-documents`. Although it's a good point, Im going to check the following  places where these might be in use:
 - The fuzzymatcher (need a change)
 - The web api (already up to date in the last PR iirc)
 - The fulltext matcher

hblanks: Gah...right. Hmm. Not sure what's right then. A few options sketched out:

A:

```
policy-docs
policy-citations
policy-test-docs
policy-test-citations
```

B:

```
policy-policy-docs
policy-citations
policy-test-policy-docs
policy-test-citations
```

C

```
policy--policy-docs
policy--citations
policy-test--policy-docs
policy-test--citations
```

I'm not really sure what's square but I'm happy with what you have. Sorry for misreading.

PR #265, opened 2019-11-01
Web search results post rename

SamDepardieu:
# Description

See #262 , as this is just a rebased version of that PR


reach/web/templates/search/policy-docs.html
```diff
@@ -39,7 +39,7 @@ <h1>Search the contents of Policy Documents</h1>
                         <div class="card">
                           {{ uparrow() }}
                           <div class="card-body">
-                            Wellcome Reach regularly loads the PDFs of WHO, MSF, UNICEF, NICE, and other
+                            Policy tool regularly loads the PDFs of WHO, MSF, UNICEF, NICE, and other
```
hblanks: "Wellcome Reach"

reach/web/views/search.py
```diff
@@ -0,0 +1,162 @@
...
+        except ConnectionError:
+            message = 'Could not join the elasticsearch server'
+            raise falcon.HTTPServiceUnavailable(description=message)
+
+        except NotFoundError:
+            message = 'No results found.'
+            return False, message
+
+        except Exception as e:
+            raise falcon.HTTPError(description=str(e))
```
hblanks: Is this also logged or reported to Sentry?

reach/web/views/search.py
```diff
@@ -0,0 +1,162 @@
...
+                    float(response['hits']['total']['value']) / params['size'])
+                )
+
+            self.context.update(params)
+            super(FulltextPage, self).render_template(
+                resp,
+                '/results/policy-docs',
+            )
+        else:
+            super(FulltextPage, self).on_get(req, resp)
```
hblanks: Looking great!

reach/web/test_api.py
```diff
@@ -1,4 +1,4 @@
-from . import api
+from .views import template
```
hblanks: This file should probably be `reach/web/views/test_template.py` with import

```python
from . import template
```

reach/web/templates/search/policy-docs.html
SamDepardieu: Ah yes, yes, had some issue with my `git stash pop`, I knew I'd miss something!

reach/web/views/search.py
SamDepardieu: That's a good question, we didn't get any exception yet, so my best answer here is: probably not. I'm going to fix that 👍

reach/web/test_api.py
SamDepardieu: Shouldn't we rather have a `tests` folder with various test files?

hblanks: Hmm. That's a good question. I seem to be letting my golang sensibilities leak into Python.

Yes, do lets have a tests folder. Sorry, too much go on the mind  I think.

SamDepardieu: No problem! Just out of curiosity, what would the Go best practice be?

hblanks: https://github.com/golang/go/tree/master/src/fmt shows a good example. All test files just have `_test.go`.

hblanks: Also, it turns out that [pytest supports both](http://doc.pytest.org/en/latest/goodpractices.html#choosing-a-test-layout-import-rules) and we use "within application code" tests in the internal repo (I think this is OK to share):

```
hblanks@ast325:~/.../wellcome/git/datalabs$ find datalabs/ -name \*test\*.py
datalabs/dimensions/test_index_researchfish.py
datalabs/dimensions/test_client.py
datalabs/epmc/test_bulk_metadata.py
datalabs/epmc/test_index_publications.py
datalabs/epmc/test_classify_publications.py
datalabs/epmc/test_bulk_open_access.py
datalabs/epmc/test_client.py
datalabs/test_pubid.py
```

We're only using `tests/` directories at this point, though. So, happy to keep things that way if it suits.

PR #283, opened 2019-11-14
Web citation search

### LINKED ISSUE #278: Reach web search use the wrong url in its pagination
## Where?
On the search results page for policy docs

## What?
The pagination links

## How?
The pagination link redirect to `/results/content` instead of `/search/policy-docs`

SamDepardieu:
# Description

This PR adds the Citation search feature to Reach web interface. It also includes a few other changes:
  - search.py: Add the search citations API and web page
 - api.py: Add the routes for the above mentioned features
 - css and Makefile: Move css files to a dedicated folder (This is to prepare a follow up PR about minifying the JS and adding a JS sorting module)
 - wsgi.py: Add a cong entry for ES indexes
 - docker-compose: Add the es indexes to use
 - Fix #278 (wrong links on pagination) as a drive by issue
 
Note that this PR only sets the basic feature for citation search. The following will come in another PR:
 - "Accordion" for multiple citations in the search results
 - Sorting on both results tables
 - JS minifying

## Type of change

- [x] :bug: Bug fix
- [x] :sparkles: New feature

# How Has This Been Tested?
Local runs & `make docker-test`
Using Docker 19.03.4 with 2CPU, 10.5GiB


reach/web/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "postcss-url": "^8.0.0",
+    "precss": "^4.0.0"
+  },
+  "devDependencies": {},
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "Wellcome Data Labs",
+  "license": "MIT"
+}
```
hblanks: @SamDepardieu  - this file doesn't seem to be pulled into the docker build...so I'm not sure how we're using it when building the web image. Am I missing something, though?

SamDepardieu: Ah yes, my bad! This is in the next PR with the JS minifying: Because we're getting quite a few modules now, instead of installing everything in `Dockerfile.node` I just copy the `package.json` file and run `npm install`. You can see it there: https://github.com/wellcometrust/reach/blob/minify-web-js/reach/web/Dockerfile.node

hblanks: That's great! Thanks for making improvements as you go.

Issue #273, opened 2019-11-08
Fix pdf extraction for all system types in pdf_parser.py


lizgzil:
After some tests failing when I ran `make test` on my mac computer @ivyleavedtoadflax realised there is an error in `reach/pdf_parser/pdf_parser.py` in which the text extraction works as desired for Linux only.

Line 89 in pdf_parser.py: 
`full_text = '\n'.join([_flatten_text(text) for text in tree.xpath('//text')])`

e.g. for the test pdf in https://github.com/wellcometrust/reach/blob/master/reach/scraper/tests/pdfs/test_pdf_multipage.pdf, my mac outputs 18 lines:
```
Test Page 1
All bold line.
Partly
bold
 line.
All italic line.
Partly
italic
 line.
Test Page 2
All bold line.
Partly
bold
 line.
All italic line.
Partly
italic
 line.
```
whereas we want to output to be 10 lines
```
Test Page 1
All bold line.
Partly bold line.
All italic line.
Partly italic line.
Test Page 2
All bold line.
Partly bold line.
All italic line.
Partly italic line.
```

ivyleavedtoadflax:
It looks like the formatted portions - i.e. bold and italic have a text attribute that is being found by `tree.xpath('\\text')`. Why that is present on mac and not linux is anyone's guess, but either `pdf2html` or `lxml` is performing differently on mac.

nsorros:
that sounds super weird and important to resolve.

jdu:
This is I think actually down to the version of poppler you have installed and not to do with a difference between Linux / macos. Homebrew on mac seems to be distributing version 0.81 of poppler which has some regressions/issues from the look of their issue list, especially around line breaks and things.

The linux distributions are around version 0.71 which don't seem to be a problem, so as long as we pin our install of poppler to <0.81 on linux we should be OK.

hblanks:
Alas. Some options I can think of here:

1. Annotating or otherwise altering the poppler-dependent tests so they run with `make docker-test` but not `make test` when run on OS X.
1. (More painfully) requiring the earlier version of poppler be installed on OS X
1. (Hard, but maybe best going forward?) Advancing the poppler version in Linux and fixing the code/tests to work with this.

Are there other ways, though? Or is one of these a good choice? What do people think?

(Aside: It's interesting to note that our current test, which passes on 0.71, is essentially a "Progression test" in that it verifies that the library we're using has not been upgraded to a version that we can't use yet.)

[...]

Issue #257, opened 2019-10-09
Exact matching is slow on local run of refparse


lizgzil:
During my work on [this PR](https://github.com/wellcometrust/policytool/pull/252) I ran locally for the 70 documents in the latest MSF scrape, using:
```
python -m policytool.refparse.refparse \
    --scraper-file "s3://datalabs-dev/reach-airflow/output/policy/parsed-pdfs/msf/parsed-pdfs-msf.json.gz" \
    --references-file "s3://datalabs-data/wellcome_publications/uber_api_publications.csv" \
    --model-file "s3://datalabs-data/reference_parser_models/reference_parser_pipeline.pkl" \
    --output-url "file://./tmp/parser-output/output_folder_name"
```
this took **504 seconds** and finds 10 doc-publication matches.

Without these changes this took **8 seconds** and found 2 doc-publication matches (reassuring also found in the 10 full text search matches).

- I tried flashtext to make it quicker, but it didn't work on looking for whole sentences in large amounts of text - it works for 1 or 2 words in text.
- I tried Whoosh to make this quicker, but it took 1774 seconds and found 7538 matches - obviously there is probably a bug somewhere, but I didn't investigate any further.
- I started to try Spacy's PhraseMatcher (`pip install spacy` and `python -m spacy download en_core_web_sm`) but stopped, this might be a good starting point

jdu:
We can potentially instrument this with cProfile to see what's eating up all the time while it's processing. I can have a look at running this next week wrapped with that to see where the time is being spent.

SamDepardieu:
Good time to try Python 3.8 `cprofile as a context manager`? 
https://docs.python.org/3/whatsnew/3.8.html#cprofile

jdu:
This looks like it's down to the use of regex in `exact_match.py` for substitutions, each time the `clean_text` method is called, it's recompiling the regular expressions. Regex isn't the fastest thing on the planet unfortunately and it may be worth exploring different options for doing these substitutions if the speed of this is a major issue.

```
		string = re.sub("\\n", " ", string)
		string = re.sub("\s{1,}", " ", string)
		string = re.sub("[^A-Za-z0-9 ]", "", string)
```

As a side note: A (relatively easy) way to diagnose these things is using `cProfile` and `kcachegrind`.

Normally you can just import cProfile into your file that you want to run and set it up like so:

```
import cProfile

def some_func():
    # Do things

if __name__ == "__main__":
[...]

PR #285, opened 2019-11-14
Add Exact Matcher

### LINKED ISSUE #202: The exact matcher is only used in reference section texts

lizgzil:
Fix #202 

This is the rebased version of https://github.com/wellcometrust/reach/pull/252

# Description

This PR adapts the code to look for exact matches in all the policy text, rather than just in the references sections.

- Adds an optional argument in `get_scraping_results` in file_manager.py which allows you to input which columns to include in the scraper output dictionary. This will mean we can include the text column needed to exact match on all the text. I did this rather than just updating the global variable `SCRAPING_COLUMNS` with 'text' since for the fuzzy match it is unneeded to store the large text data in memory.
- I create a new function `transform_scraper_text_file` in `refparse.py` to specifically transform the scraped full text for the exact match.
- Create a new dag (policy-test-exact-match) that does the exact matching (editing `exact_match_refs_operator.py` and `policy.py`).
- Adds 8 new exact matcher tests
- Changes to `test_pdf_objects` to allow for these tests to be run (or skipped) on a mac (this should be fixed properly at some point, see issue #273 )

## Type of change
- [x] :bug: Bug fix

## Running

A ran locally for the 70 documents in the latest MSF scrape, using:
```
python -m policytool.refparse.refparse \
    --scraper-file "s3://datalabs-dev/reach-airflow/output/policy/parsed-pdfs/msf/parsed-pdfs-msf.json.gz" \
    --references-file "s3://datalabs-data/wellcome_publications/uber_api_publications.csv" \
    --model-file "s3://datalabs-data/reference_parser_models/reference_parser_pipeline.pkl" \
```
this took **504 seconds** and finds 10 doc-publication matches.

Without these changes this took **8 seconds** and found 2 doc-publication matches (reassuring also found in the 10 full text search matches).

- I tried flashtext to make it quicker, but it didn't work on looking for whole sentences in large amounts of text - it works for 1 or 2 words in text.
- I tried Whoosh to make this quicker, but it took 1774 seconds and found 7538 matches - obviously there is probably a bug somewhere, but I didn't investigate any further.
- I tried Spacy's PhraseMatcher (`pip install spacy` and `python -m spacy download en_core_web_sm`) but aborted trying this out, since in airflow we dont need it, so it doesnt matter if refparse ran locally is slow

I've created an issue #257, if someone wants to pick it up this performance issue in the future.

# How Has This Been Tested?

```
>>> make docker-test
74 passed, 4 warnings in 10.56s
>>> make test
 72 passed, 2 skipped, 5 warnings in 7.34s
>>> cd reach/refparse/
>>> python -m unittest
Ran 35 tests in 0.112s

OK
```
I ran:
```
source build/virtualenv/bin/activate
eval $(./export_env.py)
docker-compose up -d
./docker_exec.sh airflow test policy-test ExactMatchRefs.msf 2019-11-8
```
it took about 3 hours to run and processed 5377194 publications and found 8132 matches. 
I noticed my results have duplicates in, fixing this might help speed things up if we delete duplicate policy docs before the exact matcher runs. I added some info about this duplication to issue #183

policy-test and policy-test-exact-match dags complete:
Screenshot 2019-11-14 at 14 24 09


reach/refparse/utils/exact_match.py
```diff
@@ -3,7 +3,7 @@
 
 class ExactMatcher:
```
jdu: Indentation is a bit wonky in this file for some reason.

See https://github.com/wellcometrust/reach/issues/257#issuecomment-555460819 about the performance of `clean_text`

hblanks: Thanks for finding this, @jdu . I've pushed a commit that fixes this and other stray tabs I found in the repo.

hblanks:
@nsorros , @lizgzil , @SamDepardieu , @jdu  - OK! I've updated so that:

1. The test DAG has both fuzzy and exact match, with appropriately limited inputs all the way down
1. The main DAG only has fuzzy match

And the DAG generation code is a little bit more modular along the way

This gets us to where at least the test DAG runs in short time, and we can move exact into the main DAG when we speed it up enough.

PTAL when you can, and we'll give it a merge.

lizgzil:
@hblanks LGTM!

hblanks:
One more update: this does increase the test DAG runtime by ~6 minutes:

gantt

@SamDepardieu , @jdu  - do let me know if you think a 15 minute runtime is acceptable. If not, we still have some options:

1. Reduce number of records matched by exact matcher in test DAG.
1. Increase parallelism (though this may be RAM intensive) by increasing `celery.worker_concurrency` from 2 to 4 in `airflow.cfg`.

SamDepardieu:
I'm happy with a 15min job

jdu:
15 minutes is fine as i've taken to running a full policy-test and then re-running individual tasks within it to test changes, etc... so it's not all that bad.

Maybe further down the road we can set up a `python -m HTTPServer` based site in the repo or something to have a more controlled/faster test target for the policy-test.

Issue #286, opened 2019-11-14
`_check_headers` for parliament issue


jdu:
Specifically, because parliament sends back `application/octet-stream` for some PDFs as the mimetype, in parliament scraper here https://github.com/wellcometrust/reach/blob/master/reach/scraper/wsf_scraping/spiders/parliament_spider.py#L90 we're specifying that we want to accept response with either the `application/octet-stream` or `application/pdf` mimetypes.

Although this will allow getting to the `save_pdf` callback, it will get stopped there as the same check is done again in `save_pdf` https://github.com/wellcometrust/reach/blob/master/reach/scraper/wsf_scraping/spiders/base_spider.py#L76 which blocks saving the PDF.

This needs to be reworked a bit and we have to be careful that if we're accepting `application/octet-stream` the resulting file could be any number of formats as it's catch-all.

The resolution for #261 has some fixes for this to be stricter about what we're actually downloading which can be worked into the other scrapers once it's approved.

jdu:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix parliament scraper not downloading `application/octet-stream` PDFs](https://trello.com/c/fP3fq9cn/903-fix-parliament-scraper-not-downloading-application-octet-stream-pdfs)

jdu:
Fix coming in #287 as the code in question here was directly impacted.

jdu:
Resolved by #261

PR #287, opened 2019-11-15
scraper/parliament: Fix an issue with file:/// protocol URLs Fix #261

### LINKED ISSUE #261: Parliament scrape is trying to handle local file:/// protocol URLs as FTP(S) download resulting in FileNotFound
It's not clear at all why, but twice now, the parliament scrape has taken to using internet-supplied input when accessing the container's filesystem. Not a great thing, really.

A couple questions that might help:

1. How does the scraper specify the path for the downloaded files? (I could have sworn we were using NamedTemporaryFiles...so not sure what's gone on here.)
1. Where does this text string come from on the Parliament website?

Sentry Issue: [AIRFLOW-POLICYTOOL-STAGING-1JJ](https://sentry.io/organizations/wellcome-data-labs/issues/1291912155/?referrer=github_integration)

```
<twisted.python.failure.Failure builtins.FileNotFoundError: [Errno 2] No such file or directory: '/C:/Users/hunterwainwrighta2/Downloads/NG_Winter_Outlook_2015_FINAL_INT (3).pdf'>
```

jdu:
# Description

This pull is to fix #261 see that issue for the specific details of the issue.

* Adds new `_is_valid_pdf` and `_is_valid_pdf_url` methods to `base_spider.py` which filter on headers as well as the URL fragment for selecting appropriate PDF files for download.

## Type of change

- [x] :bug: Bug fix

# How Has This Been Tested?

`make docker-test`

# Checklist:

### DIFF
--- a/reach/scraper/wsf_scraping/spiders/base_spider.py
+++ b/reach/scraper/wsf_scraping/spiders/base_spider.py
@@ -60,8 +60,11 @@ class BaseSpider(scrapy.Spider):
         content_type = response_headers.get('content-type', '').split(b';')[0]
         return desired_extension == content_type
 
-    def _is_valid_pdf(self, response, extension=None, mimetype=None):
+    def _is_valid_pdf(self, response):
         """ Test if a response is a PDF
+
+        We allow a pdf if it's file extension is not .pdf but
+        its content-type is explicitly `application/pdf`
 
         Args:
             response: The request response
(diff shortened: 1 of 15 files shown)


jdu:
This is WIP to get comments on implementation, there are a couple notes in addition:

1. We might want to update the tests to reflect catching the type fo scenario in #261 
2. This is explicitly requires that a download has an extension of 'pdf', do we know of any instances where a PDF is served from a route without and extension and the filename is passed as a header for instance (i.e. a route like `/files/documents/<uuid>` which has a header like `Content-Disposition: filename="some_document_2019.pdf"`

There's also an issue with filtering raised in #286 that came out of this.

reach/scraper/wsf_scraping/spiders/base_spider.py
```diff
@@ -56,6 +60,61 @@ def _check_headers(self, response_headers,
...
+            extension: The type of extension to limit to
+            mimetype: Either a **list|tuple** of or single mimetype
+                      to limit the URL to
+        """
+
+        if extension is None:
+            extension = 'pdf'
+
+        if mimetype is None:
+            mimetype = b'application/pdf'
```
SamDepardieu: Why not settings these in the header?
```
def _is_valid_pdf(self, response, extension='pdf', mimetype=b'application/pdf'):
...
```

jdu: Habit mostly, python default arguments are mutable as they're instantiated when defined and not at the point the actual method is called, so I've formed a habit of not specifying default arguments other than None to avoid polluting them down the road. 

In case you haven't seen this before and for anyone else reading this who might not have:

```
class SomeClass():
    def some_method(self, new_val, param=[]):
        param.append(new_val)
        print(param)

new_inst = SomeClass()
new_inst.some_method("HELLO")
new_inst.some_method("GOODBYE")
```

It's easy to assume that the above would output:

```
['HELLO']
['GOODBYE']
```

when it will actually output
```
['HELLO']
['HELLO', 'GOODBYE']
```

because `param` is mutable.

Instead of reasoning about every case I just built a habit out of always specifying None for keyword arguments and dealing with them explicitly in the methods body.

Can change in this case as it's not as much of an issue here, but really it just boiled down to habit...

hblanks: :+1:  yes, finding out about mutable keyword args when debugging is no fun. I'm more ok with immutable keyword args not being None but agree, you'll save yourself some pain if you use None most of the time.

SamDepardieu: That's good to know 👍

jdu:
Already done propagating changes to other scrapers, as well as some changes as the NICE site doesn't provide URLs with extensions for the PDF downloads so had to re-adjust the approach. Also updated some tests in line with these changes.

Just waiting on policytool-test to run the scrapers to make sure everything is working correctly then will update this PR.

jdu:
Refactored this a bit as the NICE scraper turned out to have URLs which don't contain a file extension for pdf. Updated some tests which didn't pass after these changes as the tests didn't specify a response with a pdf extension URL which caused them to fail the `is_vaild_pdf` and `is_valid_pdf_url` checks.

Have ran policy-test and scrapers run fine, as well as all tests passing in `make docker-test`

jdu:
Just to add to this, the parliament site emailed me back and have fixed at least the link in question which triggered the error.

PR #290, opened 2019-11-21
reach/web: disable search engine indexing

hblanks:
# Description

Extend reach/web/api.py so that it serves GET /robots.txt requests.

## Type of change

- [x] :sparkles: New feature

# How Has This Been Tested?

```
curl http://localhost:8081/robots.txt
make docker-test
```

# Checklist:

### DIFF
--- a/reach/web/api.py
+++ b/reach/web/api.py
@@ -14,7 +14,9 @@ import falcon
 from elasticsearch import Elasticsearch
 
 from reach.sentry import report_exception
-from reach.web.views import template, search
+from reach.web.views import template
+from reach.web.views import search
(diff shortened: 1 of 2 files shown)


hblanks:
> This looks fine, maybe let's keep in mind that if we start having to serve additional files out of reach (images, favicon, etc..) maybe there's a static files middleware or something that will make this easier? Tornado has a StaticFiles route but not sure if falcon has one or something similar.

Much agreed. We have this for directories within Falcon and use it for `/static` thus:

```python
    api.add_static_route('/static', conf.static_root)
```

Not so much for individual paths.

PR #293, opened 2019-11-27
Add sorting and pagination to the tables

SamDepardieu:
# Description

This PR adds sorting and pagination to the tables on Reach's web application.
 - Add List.js to the packages
 - Create some nice JS script to add sorting and pagination
 - Bit of rework on both result pages to allow pagination and sorting to happen

## Type of change

- [x] :sparkles: New feature

# How Has This Been Tested?

`make docker-test`
local runs and tests

### DIFF
--- a/reach/web/src/js/sortTables.js
+++ b/reach/web/src/js/sortTables.js
@@ -1,4 +1,5 @@
 const List = require('list.js');
+const ItemsPerPage = 25;
 
 function updateCount(htmlItem) {
     let current = 1;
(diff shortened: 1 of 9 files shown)


reach/web/src/js/sortTables.js
```diff
@@ -0,0 +1,54 @@
...
+    } else {
+    	maxEsResults = parseInt(maxEsResults.innerText);
+    }
+
+
+    if (htmlItem && htmlItem.innerText) {
+        current = parseInt(htmlItem.innerText);
+    }
+
+    let maxResults = (maxEsResults < 25 || (25 + (current - 1) * 25) > maxEsResults)? maxEsResults : 25 + (current - 1) * 25;
```
hblanks: `25` might be best as a constant

SamDepardieu:
@hblanks The table sorting lib can handle much more than that, but ES sets a soft limit to the number of results it can return (10 by default, here 1000). 

Agreed for the constant

PR #306, opened 2019-12-02
Pipeline data enhacement Fix #304

### LINKED ISSUE #304: Set up better test DAG pipeline for development
Set up a test DAG with the full pipeline using local storage to shorten development time and make it easier to test individual tasks without running the full cross-organization DAG as well as to stop hitting external sites repeatedly while testing / development.

jdu:
# Description

This started off as a bit bigger of a change but have broken into multiple changes in further pull requests.

This adds a test DAG for testing the end to end DAG process for policy using a locally running scrape target instead of a live site (so we aren't hitting lives sites during development of the pipeline).

## Type of change

- [x] :sparkles: New feature

# How Has This Been Tested?

Ran the test DAG to make sure that it works.

# Checklist:


reach/airflow/dags/test_data_flow.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def to_s3_output_dir(dag, *args):
+    """ Returns the S3 URL for any output directory for the DAG. """
+    path = '/'.join(args)
+    slug = '-'.join(args)
+    return (
+        '{{ conf.get("core", "reach_s3_prefix") }}'
+        '/output/%s/%s/%s'
+    ) % (dag.dag_id, path, slug)
```
SamDepardieu: All of this is duplicate code from the main dag. I get why you you'd want to put this in another file, but then shouldn't we refactor these into some kind of `dag/common.py` file?

test_target/target_server.py
```diff
@@ -0,0 +1,38 @@
+import os
+
+import tornado.ioloop
+import tornado.web
```
SamDepardieu: More of a question here: We're already using Falcon for the website. Idon't mind  using something different for testing purpose, but should we keep to one web framework per project for consistency?

Makefile
```diff
@@ -135,3 +135,7 @@ docker-test: docker-build
 
 .PHONY: all
 all: docker-test
+
+.PHONY: run-scrape
+run-scrape:
```
SamDepardieu: I think we should call that something like `run-test-scrape`, to avoid any confusion

reach/airflow/dags/test_data_flow.py
jdu: I had considered that, but I think overall we can address that later as a new issue for the next release to clean this up and make it easier to understand what's going on in the DAG files.

Makefile
jdu: This can actually get removed as the site is run in docker-compose now.

test_target/target_server.py
jdu: The dependency is only expressed in test_requirements.txt and not loaded into the actual application. The reason I used tornado is it has less spread of dependencies versus falcon in terms of handling additional behaviour. It's also asynchronous so that I can scale up the test server later on to handle more connections running locally by multi-threading it without the need for gunicorn or anything else in front of it.

tl;dr it's not a production dependency and it will only ever be run on local.

jdu:
Makefile directive is removed, merging

Issue #165, opened 2019-06-20
Consider fuzzy match randomness


lizgzil:
If there are multiple matches with the same similarity the code picks the first one as the match. Should we pick this randomly or is it fine to always pick the first one in the list?

The case against picking randomly is because of reducibility - if it's random we get some different results every time it is run and this might lead people to distrust the results.

The case for randomness is if there is any bias in picking the first publication from the list, for example are the earlier publications in the list from older publications?

We could use a random seed to shuffle the publications data, hence making the results reducible _and_ random, but if the publications data changes (which it will do as more publications come out) this will no longer give reducible results. Although perhaps the same can be said for picking the first in the list (will match publication X always be first in the list if there are more publications added)?

We don't currently have an estimate of how often this even happens.

nsorros:
Lately, I am leaning towards the idea of not randomness so that the results are reproducible. I know that this introduces a bias but it sounds minimal.

This could be a nice investigation for 'ethics' whether this decision, i.e. picking the first when the similarity is the same, favors particular type of matches.

hblanks:
Note you can seed the random number generator to get deterministic results.

ivyleavedtoadflax:
I think @hblanks probably has it here. Why don't we select randomly, but set the random seed to ensure that the random choice is reproducible?

ivyleavedtoadflax:
Also I presume if we retain enough decimal places on the match, this is fairly unlikely to occur?

kristinenielsen:
@lizgzil to review whether this is still an issue. If so, we can have a look at splitting it out and make it into actionable work

lizgzil:
This issue is now in lines 80-90 of https://github.com/wellcometrust/reach/blob/master/reach/airflow/tasks/fuzzy_match_refs.py
i.e.
```
res = self.es.search(
    index=self.es_index,
    body=body,
    size=1
    )
matches_count = res['hits']['total']['value']
if matches_count == 0:
    return

best_match = res['hits']['hits'][0]
```
From the internet:
"The top_hits aggregation uses the internal doc_id (in Lucene) as a tiebreak for documents with same sort values." (https://discuss.elastic.co/t/top-hits-query-with-same-score/107018)

So I _think_ this means **the top returned result is basically random** (but I'm not sure how random sorting by the internal doc_id is) **if there are multiple tie-breakers**.

@jdu and @SamDepardieu do you think my interpretation above in bold is correct? 

[...]

PR #314, opened 2019-12-06
Update fuzzy_match.py

ivyleavedtoadflax:
# Description

Updates the `FuzzyMatch` class by removing the reliance on pandas, instead using a indices and lists of dicts. This approach cuts the time taken to match 620 references against a list of 5 million publication from a ~10-15s search time to ~3-5s search time, reducing time taken from over two hours to just under one.

## Type of change

- [x] :sparkles: New feature (refactor)
- [x] :memo: Documentation update

# How Has This Been Tested?

`make # docker tests`

Also, manually labelled titles extracted from the validation for the 2019.10.8 model were matched against the 5 million publications.

# Checklist:

### DIFF
--- a/reach/refparse/tests/test_fuzzy_match.py
+++ b/reach/refparse/tests/test_fuzzy_match.py
@@ -1,119 +1,98 @@
-import unittest
-
 import pandas as pd
+import pytest
 from pandas.util.testing import assert_frame_equal
-
-from reach.refparse.settings import settings
 from reach.refparse.utils import FuzzyMatcher
 
 
-class TestFuzzyMatcherInit(unittest.TestCase):
-    def test_empty_title(self):
-        real_publications = pd.DataFrame({
-            'title': []
-        })
-        threshold = 0.75
-        with self.assertRaises(ValueError):
-            FuzzyMatcher(real_publications, threshold)
+@pytest.fixture
+def fuzzy_matcher():
+    real_publications = [
+        {'title': 'Malaria', 'pmcid': 0},
+        {'title': 'Zika', 'pmcid': 1},
+    ]
+    fuzzy_matcher = FuzzyMatcher(real_publications, similarity_threshold=0.75)
 
-    def test_no_title(self):
-        real_publications = pd.DataFrame({})
-        threshold = 0.75
-        with self.assertRaises(KeyError):
-            FuzzyMatcher(real_publications, threshold)
+    return fuzzy_matcher
 
-    def test_no_string_titles(self):
-        real_publications = pd.DataFrame({
-            'title': [1,2]
-        })
-        threshold = 0.75
-        with self.assertRaises(AttributeError):
-            FuzzyMatcher(real_publications, threshold)
+def test_empty_list():
+    real_publications = []
+    with pytest.raises(ValueError):
+        FuzzyMatcher(real_publications, similarity_threshold=0.75)
 
-    def test_init_variables(self):
-        real_publications = pd.DataFrame({
-            'title': ['Malaria', 'Zika'],
-            'pmcid': [0, 1]
-        })
-        threshold = 0
-        fuzzy_matcher = FuzzyMatcher(real_publications, threshold)
-        assert_frame_equal(
-            fuzzy_matcher.publications, real_publications
-        )
-        self.assertEqual(
-            fuzzy_matcher.similarity_threshold, threshold
-        )
-        self.assertTrue(
-            fuzzy_matcher.tfidf_matrix.size != 0
-        )
+def test_no_titles():
+    real_publications = [{}, {}, {}]
+    with pytest.raises(ValueError):
+        FuzzyMatcher(real_publications, similarity_threshold=0.75)
 
-    def test_init_variables_jsonl(self):
-        real_publications = pd.DataFrame(
-            {'title': 'Malaria', 'pmcid': 0},
-            {'title': 'Zika', 'pmcid': 1},
-        )
-        threshold = 0
-        fuzzy_matcher = FuzzyMatcher(real_publications, threshold)
-        assert_frame_equal(
-            fuzzy_matcher.publications, real_publications
-        )
-        self.assertEqual(
-            fuzzy_matcher.similarity_threshold, threshold
-        )
-        self.assertTrue(
-            fuzzy_matcher.tfidf_matrix.size != 0
-        )
+def test_no_string_titles():
+    real_publications = [{'title':1}, {'title':2}]
+    with pytest.raises(AttributeError):
+        FuzzyMatcher(real_publications, similarity_threshold=0.75)
 
-class TestFuzzyMatch(unittest.TestCase):
-    def init_fuzzy_matcher(self):
-        real_publications = pd.DataFrame({
-            'title': ['Malaria', 'Zika'],
-            'uber_id': [1, 2]
-        })
-        threshold = settings.FUZZYMATCH_SIMILARITY_THRESHOLD
-        fuzzy_matcher = FuzzyMatcher(real_publications, threshold)
+def test_init_variables():
+    real_publications = [
+        {'title': 'Malaria', 'pmcid': 0},
+        {'title': 'Zika', 'pmcid': 1},
+    ]
 
-        return fuzzy_matcher
+    threshold = 0.75
+    fuzzy_matcher = FuzzyMatcher(real_publications, threshold)
+    assert isinstance(fuzzy_matcher.publications, dict)
+    assert isinstance(fuzzy_matcher.publications[0], dict)
+    assert fuzzy_matcher.publications[0] == real_publications[0]
+    assert fuzzy_matcher.similarity_threshold == threshold
+    assert fuzzy_matcher.tfidf_matrix.size != 0
 
-    def test_empty_reference(self):
-        fuzzy_matcher = self.init_fuzzy_matcher()
-        reference = {}
-        self.assertTrue(
-            fuzzy_matcher.match(reference) is None
-        )
 
-    def test_no_match(self):
-        fuzzy_matcher = self.init_fuzzy_matcher()
-        reference = {
-            'Document id': 1,
-            'Reference id': 1,
-            'Title': 'Ebola'
-        }
-        self.assertTrue(
-            fuzzy_matcher.match(reference) is None
-        )
 
-    def test_one_match(self):
-        fuzzy_matcher = self.init_fuzzy_matcher()
-        reference = {
-            'Document id': '10',
-            'Reference id': '11',
-            'Title': 'Malaria'
-        }
-        matched_publication = fuzzy_matcher.match(reference)
-        self.assertEqual(matched_publication['Document id'], '10')
+def test_empty_reference(fuzzy_matcher):
+    reference = {}
+    assert fuzzy_matcher.match(reference) is None
 
-    def test_close_match(self):
-        real_publications = pd.DataFrame({
-            'title': ['Malaria is caused by mosquitoes'],
-            'uber_id': [1]
-        })
-        threshold = settings.FUZZYMATCH_SIMILARITY_THRESHOLD
-        fuzzy_matcher = FuzzyMatcher(real_publications, threshold)
-        reference = {
-            'Document id':  '10',
-            'Reference id': '11',
-            'Title': 'Malaria'
-        }
-        matched_publication = fuzzy_matcher.match(reference)
-        self.assertEqual(matched_publication, None)
+def test_no_match(fuzzy_matcher):
+    reference = {
+        'Document id': 1,
+        'Reference id': 1,
+        'Title': 'Ebola'
+    }
+    assert fuzzy_matcher.match(reference) is None
+
+def test_one_match(fuzzy_matcher):
+    reference = {
+        'Document id': '10',
+        'Reference id': '11',
+        'Title': 'Malaria'
+    }
+    matched_publication = fuzzy_matcher.match(reference)
+    assert matched_publication['Document id'] == '10'
+
+def test_close_match():
+    real_publications = [
+        {'title': 'Malaria is caused by mosquitos', 'pmcid': 0},
+    ]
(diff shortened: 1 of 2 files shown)


reach/refparse/utils/fuzzy_match.py
```diff
@@ -1,53 +1,108 @@
...
+                "Matched title": best_match.get("title"),
+                "Matched publication id": _str_or_null(best_match.get("uber_id")),
+                "Matched publication pmcid": _str_or_null(best_match.get("pmcid")),
+                "Matched publication pmid": _str_or_null(best_match.get("pmid")),
+                "Matched publication doi": _str_or_null(best_match.get("doi")),
+                "Similarity": best_similarity,
+                "Match algorithm": "Fuzzy match",
             }
+
+def _str_or_null(docid):
```
nsorros: why is this needed?

reach/refparse/utils/fuzzy_match.py
```diff
@@ -1,53 +1,108 @@
...
         if best_similarity > self.similarity_threshold:
             return {
-                'Document id': reference['Document id'],
-                'Reference id': reference['Reference id'],
-                'Extracted title': reference['Title'],
-                'Matched title': best_match['title'],
-                'Matched publication id': best_match['uber_id'],
-                'Similarity': best_similarity,
-                'Match algorithm': 'Fuzzy match'
+                "Document id": _str_or_null(reference.get("Document id")),
```
nsorros: a good question to ask is why would all these information be missing if there is a match? and is that a behaviour we want?

reach/refparse/utils/fuzzy_match.py
```diff
@@ -1,53 +1,108 @@
...
+            return None
+
+        if len(reference.get("Title")) < self.title_length_threshold:
+            return None
 
         retrieved_publications = self.search_publications(reference)
 
         best_match = retrieved_publications.iloc[0]
-        best_similarity = best_match['similarity']
+        best_similarity = best_match.get("similarity")
```
nsorros: isn't there a problem if there is no similarity?

reach/refparse/utils/fuzzy_match.py
```diff
@@ -1,53 +1,108 @@
...
+
+        Args:
+            publications(list): List of dicts containing publication info as
+                above, or a pandas.DataFrame.
+            similarity_threshold(float): Minimum allowable similarity
+                threshold. If not results are found to have a similarity
+                higher than this, then nothing is returned.
+            title_length_threhold(float): Minimum allowed length of title.
+                Less than this threshold will return no results.
+        """
```
nsorros: super helpful, thanks for adding

reach/refparse/utils/fuzzy_match.py
```diff
@@ -1,53 +1,108 @@
...
-        if len(reference['Title']) < self.title_length_threshold:
-            return
+            return None
+
+        if len(reference.get("Title")) < self.title_length_threshold:
+            return None
 
         retrieved_publications = self.search_publications(reference)
 
         best_match = retrieved_publications.iloc[0]
```
lizgzil: FYI if you haven't seen - there is an issue open about this: https://github.com/wellcometrust/reach/issues/165 Please contribute to the discussion if you have any opinions about it

reach/refparse/utils/fuzzy_match.py
```diff
@@ -1,53 +1,108 @@
...
+                "Matched publication id": _str_or_null(best_match.get("uber_id")),
+                "Matched publication pmcid": _str_or_null(best_match.get("pmcid")),
+                "Matched publication pmid": _str_or_null(best_match.get("pmid")),
+                "Matched publication doi": _str_or_null(best_match.get("doi")),
+                "Similarity": best_similarity,
+                "Match algorithm": "Fuzzy match",
             }
+
+def _str_or_null(docid):
+    if docid:
```
lizgzil: So if docid = None, then `_str_or_null` will return the string "None"? Just wondering whether this will mess up anything down the line, might be worth keeping it as None. Or is there a reason to convert it to a string?

reach/refparse/utils/fuzzy_match.py
ivyleavedtoadflax: I was having some issues where the pmid (which looks like an int) was being converted to string as: `4576358.0`. When the data is coerced to `pandas.DataFrame` the column has type `float` applied to it. I could fix it in pandas, but because it wasn't entirely clear to me whether the input `publications` should be a dataframe or a list of dicts, I added a function here.

reach/refparse/utils/fuzzy_match.py
ivyleavedtoadflax: Good question. I probably need to re-add the document id to the manually tagged data.

reach/refparse/utils/fuzzy_match.py
ivyleavedtoadflax: Yep you're right, this is probably not required

reach/refparse/utils/fuzzy_match.py
```diff
@@ -1,53 +1,108 @@
...
+        ]
+        ```
+
+        Args:
+            publications(list): List of dicts containing publication info as
+                above, or a pandas.DataFrame.
+            similarity_threshold(float): Minimum allowable similarity
+                threshold. If not results are found to have a similarity
+                higher than this, then nothing is returned.
+            title_length_threhold(float): Minimum allowed length of title.
```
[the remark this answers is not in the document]
ivyleavedtoadflax: yep, good spot!

reach/refparse/utils/fuzzy_match.py
ivyleavedtoadflax: The intended function is: if the `id` is a `float`, then convert to `int` before returning as `str`.

This is a good question. This partly comes back to the `list` of `dict`s/`pd.DataFrame` question - what should the input of `publications` be. If the latter, then we can probably just sanitize the column types in the DataFrame itself.

reach/refparse/utils/fuzzy_match.py
ivyleavedtoadflax: Also on this @nsorros, I have set the threshold as a `float` (0:1), but in various places I have seen it as an integer. Which is the correct?

reach/refparse/tests/test_fuzzy_match.py
```diff
@@ -1,38 +1,39 @@
...
+from reach.refparse.utils import FuzzyMatcher
 
 
 class TestFuzzyMatcherInit(unittest.TestCase):
     def test_empty_title(self):
         real_publications = pd.DataFrame({
             'title': []
         })
-        threshold = 75
+        threshold = 0.75
```
ivyleavedtoadflax: Is this right  @lizgzil @nsorros? Should these be decimals or integers?

reach/refparse/utils/fuzzy_match.py
nsorros: that's a good question, let me have a look

reach/refparse/tests/test_fuzzy_match.py
nsorros: i think that is right, it seems like a "bug"

nsorros: https://github.com/wellcometrust/reach/blob/df709550f9bdcb7653cc0912e13047d9768f6604/reach/refparse/settings.py#L10

reach/refparse/utils/fuzzy_match.py
nsorros: https://github.com/wellcometrust/reach/blob/df709550f9bdcb7653cc0912e13047d9768f6604/reach/refparse/settings.py#L10

reach/refparse/utils/fuzzy_match.py
nsorros: This is the policy id so without it you will not be able to match, or you could but you would need to map the titles back to the ids.

reach/refparse/utils/fuzzy_match.py
```diff
@@ -37,38 +37,71 @@ def __init__(
...
                 higher than this, then nothing is returned.
-            title_length_threhold(float): Minimum allowed length of title.
+            title_length_threhold(int): Minimum allowed length of title.
                 Less than this threshold will return no results.
         """
-        # Filter out any publications that don't have titles
+        # Filter out any publications that don't have titles assuming that the
+        # input is a jsonl.
 
-        self.publications = pd.DataFrame(publications)
```
nsorros: so these changes are mainly to eliminate the need for pandas?

reach/refparse/utils/fuzzy_match.py
```diff
@@ -37,38 +37,71 @@ def __init__(
...
+            doc.update({"similarity": similarity_vector[key]})
+
+        # Convert to a list, again checking that sorting is correct
+
+        retrieved_publications = [value for key, value in retrieved_publications.items()]
+
+
+        retrieved_publications = sorted(retrieved_publications, key=lambda x: x['similarity'], reverse=True)
+
+        return retrieved_publications
```
nsorros: i am not 100% convinced that this is a better implementation. it substitutes 6 lines for 27. what is the main motivation behind it?

ivyleavedtoadflax: The old implementation was making a deepcopy of the publications dataframe every time it did a match. This dataframe has 5 million rows, so this step considerably slows the process. On my machine it locks up >20GB of RAM and maxes out one core (it's not parralelisable) for about 11-15 seconds per query. On the current task I of matching manual labels, that is ~620 * ~13 seconds = 2.23 hours. This new implementation takes ~620 * ~4 seconds = 0.688 hours. Speed is the only motivation!

reach/refparse/utils/fuzzy_match.py
ivyleavedtoadflax: See the following comment. These changes are for speed, and the existing implementation was inefficient because it created deepcopies of a very large dataframe, and it also executed repeated searches across the whole dataframe. By relying on indexes, I've managed to eliminate the slow search.

ivyleavedtoadflax:
Implemeted in https://github.com/wellcometrust/reach/pull/314/commits/53d98ac59f032ab51164ba0a3678a99cac070f73.

ivyleavedtoadflax:
> I think we need to have some fields as required not everything is optional for fuzzy matching to make sense

Implemeted in 53d98ac.

reach/refparse/utils/fuzzy_match.py
```diff
@@ -1,53 +1,136 @@
...
+
+        # Cycle through just the similar docs, not the entire list of
+        # publications.
+
+        retrieved_publications = {i:self.publications[i] for i in similar_doc_indices}
+
+        # Finally append the similarity to the similar docs based on its index
+
+        for key, doc in retrieved_publications.items():
+            doc.update({"similarity": similarity_vector[key]})
```
nsorros: I am not sure why you need the intermediate key i everywhere. can you not just zip publications and similarities?

```
for similarity, publication in zip(sorted_similarities, publications)
    publication['similarity'] = similarity
```

reach/refparse/utils/fuzzy_match.py
nsorros: Again, why would that be a get? we need a similarity, otherwise fuzzy matching does not make sense.

reach/refparse/utils/fuzzy_match.py
```diff
@@ -1,53 +1,136 @@
...
-                'Extracted title': reference['Title'],
-                'Matched title': best_match['title'],
-                'Matched publication id': best_match['uber_id'],
-                'Similarity': best_similarity,
-                'Match algorithm': 'Fuzzy match'
+                "Document id": reference["Document id"],
+                "Reference id": reference["Reference id"],
+                "Extracted title": reference["Title"],
+                "Matched title": best_match["title"],
+                "Matched publication id": best_match.get("uber_id"),
```
nsorros: i think you need one of those ids, otherwise what did you match with?

ivyleavedtoadflax: True, but we don't know which ones are going to be missing

reach/refparse/utils/fuzzy_match.py
ivyleavedtoadflax: We could just zip them together, but that will mean that we need to traverse the entire publications list, and that is what I am trying to avoid. This way we can more efficiently extract the indices of interest here: `retrieved_publications = {i:self.publications[i] for i in similar_doc_indices}` - we only traverse the `similar_doc_indices`, not the entire `self.publications`.

reach/refparse/utils/fuzzy_match.py
ivyleavedtoadflax: yep, sorry meant to change this. It could be a []

reach/refparse/utils/fuzzy_match.py
nsorros: one of them needs to be there, it is the unique identifier of the publication whatever that is. what i am trying to say is that the tool assume you pass in a list of publications with a unique identifier so you need to maintain that functionality

ivyleavedtoadflax: At present one or more can be absent from the publications data, so I would rather get all three, accept that some night be missing, and then work on creating a better if (maybe a hash of all three IDs) in the next step.

nsorros: Which publication data are you referring to?

reach/refparse/utils/fuzzy_match.py
```diff
@@ -1,53 +1,136 @@
-import numpy as np
 import logging
 
+import numpy as np
+
```
nsorros: remove newline

reach/refparse/utils/fuzzy_match.py
nsorros: 👍

reach/refparse/utils/fuzzy_match.py
nsorros: cool, good to go when you do the change

Issue #48, opened 2019-02-04
Same document different file hash


nsorros:
Just realized that some documents we scrape have a different file hash even though they represent the same document.

If our intention is not to redownload the same document, this is one problem, but even if we do not care about this, during analysis there needs to be a field which is unique per document in order for the conclusions to be accurate. The solution could be to document that document_uri is the unique key that people should use of they want to analyze the document or it could be to alter our implementation of file hash accordingly or it can be something else.

lizgzil:
An example of this is the policy document 

http://www.fao.org/3/I9553EN/i9553en.pdf

which has the document ID 

818aff942d9813d338fe31828ee9452a

 and also

14748f5b61ec161bc226354edbeee7f1

ivyleavedtoadflax:
I think this issue has reared its head again. 

The gold annotated data that I am using to evaluate reach has 81 annotated documents from a scrape done by Reach on 2019-10-8. Not all of these document ids exist in more recent scrapes however:

```
|Date of scrape |Gold documents found|
|---------------|--------------------|
|   Dec/Jan     |        10/81       |
|   2020-01-13  |        7/81        |
|   2020-01-13  |        6/81        |
|   2020-01-19  |        4/81        |
```

The problem seems to be getting worse with time. Is there any reason why taking an md5 hash of the entire document (https://github.com/wellcometrust/reach/blob/d66379b2b40ea6d01e53c4667f98db2cf2ca2896/reach/scraper/wsf_scraping/file_system.py#L11) could yield different results on subsequent scrapes @SamDepardieu @jdu ?

jdu:
It's possible it could change, but not sure the cases where this would be the case apply here. For instance if there was a process on their end which opened the file and saved it, even without changing the content of the PDF, there's a trailer in the file which contains a CreationDate/ModDate value, which if those changed it might completely change the MD5 hash output. 

Are we sure that the hashes changed and not that that the documents just aren't there a tall?

jdu:
If you have the original file from your gold data you could calc a hash of it, pull the same file from S3 and calc the hash on it and if they differ you could compare the file in a diff viewer to see if there are any differences.

ivyleavedtoadflax:
> It's possible it could change, but not sure the cases where this would be the case apply here. For instance if there was a process on their end which opened the file and saved it, even without changing the content of the PDF, there's a trailer in the file which contains a CreationDate/ModDate value, which if those changed it might completely change the MD5 hash output.
> 
[...]

PR #317, opened 2019-12-19
Incorporate neural network based reference splitter

ivyleavedtoadflax:
# Description

Switches out the existing `reference_splitter.split_section` function for the `deep_reference_parser.split_section` class. 

## Type of change

- [x] :sparkles: Adds neural network based reference splitter.

# How Has This Been Tested?

`make virtualenv test docker-test`
Ran policy-test dag in dev environment

# Checklist:

- [x] New and existing unit tests pass locally with my changes. Note that I have changed some tests to xfail as they depend on the probabilistic results of the model, which currently fails in these cases, but may not in the future.


ivyleavedtoadflax:
@jdu @SamDepardieu Would be good if you could take a look at this in regards to how reach gets access to the `deep_reference_parser` wheel. Thanks!

ivyleavedtoadflax:
Note that local tests pass. But docker tests fail with the module `deep_reference_parser` unable to be found (as expected).

reach/refparse/algo_evaluation/evaluate_split_section.py
```diff
@@ -69,8 +69,9 @@ def evaluate_split_section(evaluate_split_section_data, threshold):
     """
 
     actual_pred_num_references = []
+    section_splitter = SplitSection()
     for references_section in evaluate_split_section_data['Reference section']:
-        references = split_section(references_section)
+        references = section_splitter.split(references_section)
         actual_pred_num_references.append({
             "Predicted references" : references,
```
jdu: Please use snake case for dict keys

ivyleavedtoadflax: Tthis is a class method, not a dict key.

ivyleavedtoadflax: Agree - but lets address that in a future PR. I don't want to go down the rabbit hole of chasing where all these keys end up.

Makefile
```diff
@@ -7,7 +7,15 @@ ECR_IMAGE := 160358319781.dkr.ecr.eu-west-1.amazonaws.com/$(IMAGE)
...
 
-REFERENCE_SPLITTER_URL := https://datalabs-public.s3.eu-west-2.amazonaws.com/references_splitter/reference_splitter-2019.8.0-py3-none-any.whl
+#
+# Wheel for deep_reference_parser
+#
+
+DEEP_REFERENCE_PARSER_WHEEL := deep_reference_parser-2019.12.1-py3-none-any.whl
+DEEP_REFERENCE_PARSER_URL := https://datalabs-public.s3.eu-west-2.amazonaws.com/deep_reference_parser/$(DEEP_REFERENCE_PARSER_WHEEL)
+
+SPACY_MODEL_URL := https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-2.1.0/en_core_web_sm-2.1.0.tar.gz#egg=en_core_web_sm==2.1.0
```
nsorros: aren't these two installed in the setup.py of the deep reference parser?

ivyleavedtoadflax:
[Add neural splitter model to Reach](https://trello.com/c/opwNiSlD/923-add-neural-splitter-model-to-reach)

Makefile
```diff
@@ -79,12 +87,13 @@ docker-push: docker-test
...
 #
 
-$(VIRTUALENV)/.installed: requirements.txt test_requirements.txt
+$(VIRTUALENV)/.installed: requirements.txt test_requirements.txt $(REFERENCE_SPLITTER_WHEEL)
 	@if [ -d $(VIRTUALENV) ]; then rm -rf $(VIRTUALENV); fi
 	@mkdir -p $(VIRTUALENV)
 	virtualenv --python python3 $(VIRTUALENV)
 	AIRFLOW_GPL_UNIDECODE=yes $(VIRTUALENV)/bin/pip3 install -r requirements.txt
 	$(VIRTUALENV)/bin/pip3 install -r test_requirements.txt
+	$(VIRTUALENV)/bin/pip3 install $(DEEP_REFERENCE_PARSER_URL)
```
nsorros: why do you need this line since it is added on requierements?

nsorros:
What is this waiting to integrate?

Makefile
ivyleavedtoadflax: They need to be installed here if you created a local virtualenv. If you install from a wheel like in the docker container, it is fine.

Makefile
ivyleavedtoadflax: Slightly stretching my memory now, but I think it was because we can't add these dependencies to unpinned requirements, so they don't get propagated into requirements.txt

ivyleavedtoadflax:
> What is this waiting to integrate?

It was ready to go and waiting for approval now that we know it performs better than the existing model. However, I was planning to get the end-to-end evaluation in airflow running first so that we can be absolutely sure #341. Some changes that @jdu has made to keys may now need to be reconciled here, but I've not looked into it yet. There is some delay now on #341 due to issues that I am experiencing on the fuzzy matcher. Will add an issue about this soon, once I have investigated further.

Makefile
nsorros: But few lines below there are. I don't this causes any problems but might be redundant.

ivyleavedtoadflax:
This is now pending the deploy of #368 which will give us evaluation metrics for the existing pipeline. We can then merge this and re-run to get the metrics for the deep reference parser.

ivyleavedtoadflax:
This is blocked pending a resolution to https://github.com/wellcometrust/reach/issues/48

ivyleavedtoadflax:
We decided to merge this based on a results from a manually run evaluation which showed an improvement over the existing reference splitter.

nsorros:
Epic 🥇 Nice.

PR #321, opened 2020-01-08
Fix server memory crashes by paginating the query to ES

### LINKED ISSUE #294: Elasticsearch only returns a limited amount of data without pagination on search
Currently, we retrieve data from Elasticsearch in one `multi-match` query, of size 1000. This causes the Web app search function to be virtually limited to the first 1000 results.

The best work around this issue I can see would be to make the ES calls in the Javascript and to use Elasticsearch's pagination, allowing us to get more than the max_query_size of data. One catch with this would be that we need then to implement pagination and sorting ourselves, because it will then rely on ES order_by and pagination.

SamDepardieu:
# Description

Fix #294 

This PR changes the way we get results from Elasticsearch. Previously we were querying a batch of 1000 item and then got it paginated by the frontend. This PR uses calls to Reach API to get paginated and sorted results and rebuild the results in the table from that.

In addition, it also includes the new fields parsed in Airflow and some preparatory changes for the Citations results grouping.

(Sorry this is kind of big, but you can safely ignore the HTML files)

## Type of change

- [x] :sparkles: New feature

# How Has This Been Tested?

`make docker-test`
Local runs on Docker with 2CPU, 10.5GB memory


reach/web/src/js/resultsCommon.js
```diff
@@ -0,0 +1,125 @@
...
+    let xhr = new XMLHttpRequest();
+
+    let url = `/api/search/${type}`
+              + `?term=${body.term}`
+              + `&fields=${body.fields}`
+              + `&size=${body.size}`
+              + `&sort=${body.sort}`
+              + `&order=${body.order}`
+              + `&page=${body.page + 1}`;
+    console.log(url);
```
jdu: !!

reach/elastic/common.py
```diff
@@ -134,6 +135,21 @@ def count_es(es, index_name):
...
+    """Check is a mapping is active on given index and return it.
+    Else return False. """
+    logging.info('check_mapping: checking current mapping status')
+    try:
+        mapping = es.indices.get_mapping(index=index_name)
+        return mapping if mapping else False
+
+    except (TransportError, NotFoundError) as e:
+        if e.error == "index_not_found_exception":
+            pass
```
jdu: Possibly return an Explicit False here for clarity instead of `pass`

reach/web/views/search.py
```diff
@@ -57,6 +65,74 @@ def _search_es(es, es_index, params, explain=False):
...
+        try:
+            fields = params.get('fields', '').split(',')
+            size = params.get('size', 100)
+            es.cluster.health(wait_for_status='yellow')
+
+            es_body = {
+                'size': int(size),
+                'query': {
+                    'multi_match': {
+                        'query': params.get('term'),
```
jdu: What happens if this is None or ""?

jdu:
Will give running this a try tomorrow and then finish review.

reach/web/src/js/resultsCommon.js
SamDepardieu: Whoops

reach/elastic/common.py
SamDepardieu: 👍

reach/web/views/search.py
SamDepardieu: ES default behaviour is to return nothing if no term is specified (Empty query = no term to analyse for search = no result, IIRC)

jdu: Might be worth checking the param and short-circuiting before the query to avoid round-tripping to elastic unnecessarily.

SamDepardieu: Fair enough, I'll do that this morning

jdu:
Testing this shortly.

PR #341, opened 2020-01-16
Add Reach end-to-end evaluation to policy dag

ivyleavedtoadflax:
# Description

Adds a series of tasks required to complete an end-to-end evaluation of reach by comparing the results from the policy dag with manually labelled validation (gold) data (https://trello.com/c/BD44uSl6/922-add-end-to-end-evaluation-to-airflow-policy-dag).

Two tasks are added to the existing `policy` and `policy-test` dags yielding the following dags:

`policy`:
image

`policy-test`:
image

An additional dag (`match_annotated_titles`) is also added to ensure that the same logic for matching references found by reach can also be used for manual labelling. The output of this dag would then be manually verified, and uploaded to s3 where it can be accessed by the `policy` and `policy-test` by `evaluator.EvaluateOperator()`. The `match_annotated_titles` dag looks like:

image

NOTE: There remains an issue with fuzzy matching the gold labelled data against EPMC, which is hard to debug without having clean data. I will address this in a future PR, and it won't affect the functionality of this PR, just the accuracy.

## Type of change

- [x] :sparkles: New feature

# How Has This Been Tested?

* `make docker-test`
* Each airflow component tested independently with `airflow test`
* Run policy-test dag on local machine

# Checklist:

### DIFF
--- a/reach/airflow/dags/policy.py
+++ b/reach/airflow/dags/policy.py
@@ -188,7 +188,7 @@ def org_fuzzy_match(dag, organisation, item_limits, parsePdf, esIndexPublication
     return esIndexFuzzyMatched, fuzzyMatchRefs
 
 
-def evaluate_matches(dag, organisation):
+def evaluate_matches(dag, fuzzyMatchRefs):
     """
     Evaluate matches against a manually labelled gold dataset
     """
@@ -216,7 +216,7 @@ def evaluate_matches(dag, organisation):
         task_id='EvaluateFuzzyMatchGoldRefs',
         es_hosts=get_es_hosts(),
         src_s3_key=extractedGoldRefs.dst_s3_key,
-        organisation=organisation,
+        organisation='gold',
         dst_s3_key=to_s3_output(
             dag, 'evaluation', 'fuzzy-matched-gold-refs', '.json.gz'),
         es_index='-'.join([dag.dag_id, 'epmc', 'metadata']),
@@ -232,8 +232,7 @@ def evaluate_matches(dag, organisation):
         dag=dag,
     )
 
-    combineReachFuzzyMatchRefs >> extractedGoldRefs >> fuzzyMatchGoldRefs >> evaluateRefs
-    return evaluateRefs
+    fuzzyMatchRefs >> combineReachFuzzyMatchRefs >> extractedGoldRefs >> fuzzyMatchGoldRefs >> evaluateRefs
 
 
 def create_org_pipeline_fuzzy_match(dag, organisation, item_limits, spider_years,
@@ -249,16 +248,15 @@ def create_org_pipeline_fuzzy_match(dag, organisation, item_limits, spider_years
     """
 
     parsePdf = org_refs(dag, organisation, item_limits, spider_years)
-    esIndexFuzzyMatched = org_fuzzy_match(
+    esIndexFuzzyMatched, fuzzyMatchRefs = org_fuzzy_match(
         dag, organisation, item_limits, parsePdf, esIndexPublications)
-    evaluatedMatches = evaluate_matches(dag, organisation)
 
-    return parsePdf, esIndexFuzzyMatched
+    return parsePdf, esIndexFuzzyMatched, fuzzyMatchRefs
 
 def create_org_pipeline_exact_match(dag, organisation, item_limits,
                                     spider_years, parsePdf=None,
                                     esIndexFullTexts=None):
-    """ Creates all tasks needed for producing exact matched citatinos
+    """ Creates all tasks needed for producing exact matched citations
     for a single organisation::
 
         Spider -> ParsePdf
@@ -309,13 +307,21 @@ def create_dag_fuzzy_match(dag_id, default_args, spider_years,
         schedule_interval='0 0 * * 0,3'
     )
     esIndexPublications = es_index_publications(dag, item_limits)
+    fuzzyMatchRefsOperators = []
     for organisation in ORGANISATIONS:
-        create_org_pipeline_fuzzy_match(
+        _, _, fuzzyMatchRefs = create_org_pipeline_fuzzy_match(
             dag,
             organisation,
             item_limits,
             spider_years,
             esIndexPublications)
+        fuzzyMatchRefsOperators.append(fuzzyMatchRefs)
+
+    # Run the evaluation setting all the fuzzMatchRefsOperators as upstream
+    # depdendencies.
+
+    evaluate_matches(dag, fuzzyMatchRefsOperators)
+
     return dag
 
 
@@ -338,13 +344,16 @@ def create_dag_all_match(dag_id, default_args, spider_years,
     )
     epmc_limits = ItemLimits(None, None)
     esIndexPublications = es_index_publications(dag, epmc_limits)
+    fuzzyMatchRefsOperators = []
     for organisation in ORGANISATIONS:
-        parsePdf, _ = create_org_pipeline_fuzzy_match(
+        parsePdf, _, fuzzyMatchRefs = create_org_pipeline_fuzzy_match(
             dag,
             organisation,
             item_limits,
             spider_years,
             esIndexPublications)
+
+        fuzzyMatchRefsOperators.append(fuzzyMatchRefs)
 
         create_org_pipeline_exact_match(
             dag,
@@ -353,6 +362,11 @@ def create_dag_all_match(dag_id, default_args, spider_years,
             spider_years,
             parsePdf=parsePdf,
             esIndexFullTexts=None)
+
+    # Run the evaluation setting all the fuzzMatchRefsOperators as upstream
+    # depdendencies.
+
+    evaluate_matches(dag, fuzzyMatchRefsOperators)
 
     return dag
 
--- a/reach/airflow/dags/policy.py
+++ b/reach/airflow/dags/policy.py
@@ -11,6 +11,7 @@ from reach.airflow.tasks import es_index_fuzzy_matched
 from reach.airflow.tasks import exact_match_refs_operator
 from reach.airflow.tasks import fuzzy_match_refs
 from reach.airflow.tasks import policy_name_normalizer
+from reach.airflow.tasks import evaluator
 
 from reach.airflow.tasks.spider_operator import SpiderOperator
 from reach.airflow.tasks.extract_refs_operator import ExtractRefsOperator
@@ -32,12 +33,14 @@ DEFAULT_ARGS = {
     'retry_delay': datetime.timedelta(minutes=5),
 }
 
+GOLD_DATA = "s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid_TITLE.jsonl.gz"
+VALIDATION_DATA = "s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid.jsonl.gz"
+
 ItemLimits = namedtuple('ItemLimits', ('spiders', 'index'))
 
 #
 # Configuration & paths
 #
-
 
 def verify_s3_prefix():
     reach_s3_prefix = conf.get("core", "reach_s3_prefix")
@@ -73,6 +76,14 @@ def to_s3_output_dir(dag, *args):
         '{{ conf.get("core", "reach_s3_prefix") }}'
         '/output/%s/%s/%s'
     ) % (dag.dag_id, path, slug)
+
+def to_s3_output_dir_short(dag, *args):
+    """ Returns the S3 URL for any output directory for the DAG. """
+    path = '/'.join(args)
+    return (
+        '{{ conf.get("core", "reach_s3_prefix") }}'
+        '/output/%s/%s'
+    ) % (dag.dag_id, path)
 
 
 EPMC_METADATA_KEY = '/'.join([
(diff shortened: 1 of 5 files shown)


reach/airflow/tasks/evaluate_operator.py
```diff
@@ -0,0 +1,81 @@
...
+        self.aws_conn_id = aws_conn_id
+
+    @report_exception
+    def execute(self, context):
+
+        s3 = WellcomeS3Hook(aws_conn_id=self.aws_conn_id)
+
+        results = []
+
+        with tempfile.TemporaryFile(mode='rb+') as tf:
```
nsorros: you can wrap this block to a function that yields results to avoid the saving them in the intermediate variable results and write as you go

```
def yield_results():
    with tempfile.TemporaryFile(mode='rb+') as tf:
            # Get the precursor file (the fuzzyMatchedRefs)

            key = s3.get_key(self.src_s3_key)
            key.download_fileobj(tf)
            tf.seek(0)

            with gzip.GzipFile(mode='rb', fileobj=tf) as f:

                # Open the gzipped json file

                for fuzzy_matched_ref in f:
                    yield fuzzy_matched_ref

 # Write the results to S3
  with tempfile.NamedTemporaryFile(mode='wb') as output_raw_f:
        with gzip.GzipFile(mode='wb', fileobj=output_raw_f) as output_f:
           for item in yield_results:
                #output_f.write(item.encode("utf-8"))
                 output_f.write(item)
                  output_f.write(b"\n")
```

ivyleavedtoadflax: Nice. Very WIP!

ivyleavedtoadflax:
[Add end-to-end evaluation to airflow policy dag](https://trello.com/c/BD44uSl6/922-add-end-to-end-evaluation-to-airflow-policy-dag)

reach/airflow/dags/policy.py
```diff
@@ -17,6 +17,7 @@
 from reach.airflow.tasks.parse_pdf_operator import ParsePdfOperator
 from reach.airflow.tasks.evaluate_operator import EvaluateOperator
 from reach.airflow.tasks.evaluate_extract_refs_from_gold_data import ExtractRefsFromGoldDataOperator
+from reach.airflow.tasks.evaluate_combine_reach_fuzzy_matches import EvaluateCombineReachFuzzyMatches
```
nsorros: maybe it is worth keeping the convention of adding operator in the end

ivyleavedtoadflax: its a bit patchily applied. But agree. will do

reach/airflow/dags/policy.py
```diff
@@ -32,13 +33,15 @@
     'retry_delay': datetime.timedelta(minutes=5),
 }
 
+GOLD_DATA = "s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid_TITLE.jsonl.gz"
+VALIDATION_DATA = "s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid.jsonl.gz"
```
nsorros: remind me the difference between validation and gold?

reach/airflow/tasks/evaluator.py
```diff
@@ -0,0 +1,275 @@
...
+    template_fields = (
+        'organisations',
+        'src_s3_dir_key',
+        'dst_s3_key',
+    )
+
+    @apply_defaults
+    def __init__(self, organisations, src_s3_dir_key, dst_s3_key,
+                 aws_conn_id="aws_default", *args, **kwargs):
+        """
```
nsorros: Fill this

reach/airflow/tasks/evaluator.py
```diff
@@ -0,0 +1,275 @@
...
+def _get_fuzzy_matches(s3, src_s3_dir_key, organisations):
+    """Get all the reach fuzzy matches from all organisations
+    and combine into a json.gz.
+    """
+    task = os.path.split(src_s3_dir_key)[-1]
+
+    fuzzy_matches = []
+
+    for org in organisations:
+        src_s3_key = f"{src_s3_dir_key}/{org}/{task}-{org}.json.gz"
```
nsorros: i wonder whether this should be better passed as an arg? what if the dag changes to save in a different location, this will break and the person doing the change will not be aware.

reach/airflow/tasks/evaluator.py
```diff
@@ -0,0 +1,275 @@
...
+        with tempfile.TemporaryFile(mode='rb+') as tf:
+            valid_key = s3.get_key(src_s3_key)
+            valid_key.download_fileobj(tf)
+            tf.seek(0)
+
+            fuzzy_matches.extend(list(_yield_jsonl_from_gzip(tf)))
```
nsorros: you can also make this more memory efficient if you yield and write refs line by line

reach/airflow/tasks/evaluator.py
```diff
@@ -0,0 +1,275 @@
...
+        'valid_s3_key',
+        'gold_s3_key',
+        'dst_s3_key',
+    )
+
+    @apply_defaults
+    def __init__(self, valid_s3_key, gold_s3_key, dst_s3_key,
+                 aws_conn_id="aws_default", *args, **kwargs):
+        """
+        """
```
nsorros: fill

reach/airflow/dags/policy.py
```diff
@@ -174,7 +185,54 @@ def org_fuzzy_match(dag, organisation, item_limits, parsePdf, esIndexPublication
...
+    extractedGoldRefs = evaluator.ExtractRefsFromGoldDataOperator(
+        task_id='EvaluateExtractRefsFromGoldData',
+        valid_s3_key=VALIDATION_DATA,
+        gold_s3_key=GOLD_DATA,
+        dst_s3_key=to_s3_output(
+            dag, 'evaluation', 'extracted-gold-refs', '.json.gz'),
+        dag=dag,
+    )
+
+    fuzzyMatchGoldRefs = fuzzy_match_refs.FuzzyMatchRefsOperator(
```
nsorros: what is this for? is it to retrieve the pub id from the annotated titles? if so, then this might not be ideal because it might introduce errors which we want to measure with the gold dataset

reach/airflow/dags/policy.py
ivyleavedtoadflax: Yes good point, perhaps need to tidy up the terminology. The valuation set is from the deep reference parser model. Literally the same validation set used in the model. The gold data set is the same validation set but with titles labelled into it. In time I don't think this step should be in the dag, but right now re-matching these two datasets is a necessary step as the gold set loses all is metadata coming out of Prodigy. In fact, I would like to move that whole deep reference parser data prep pipe line into airflow in the not too distant future.

reach/airflow/tasks/evaluator.py
ivyleavedtoadflax: Fair point

reach/airflow/tasks/evaluator.py
ivyleavedtoadflax: Ah instead of collecting in a list first?

reach/airflow/dags/policy.py
ivyleavedtoadflax: Prior to this step, we have data with {doc_id, reference title}. This step matches against EPMC in the same way that we do in reach, then returns a matched object just like elsewhere in reach, do now we essentially have {pmcid, title, doc_id}. We need all three in order to evaluate reach.

reach/airflow/dags/policy.py
nsorros: ok so basically there should be one but there are two to get back the metadata. maybe add a todo and a comment explaining validation.

ivyleavedtoadflax: I think they're is a comment already in the operator. I will check.

reach/airflow/tasks/evaluator.py
ivyleavedtoadflax: Will remove. There is a class docstring.

reach/airflow/tasks/evaluator.py
ivyleavedtoadflax: Will remove. There is a class docstring.

reach/airflow/tasks/evaluator.py
nsorros: yes

reach/airflow/tasks/evaluator.py
ivyleavedtoadflax: The base is passed as an arg, but its quite hard to pass it as an arg here as it would need to be _n_ args where _n_ is the number of providers. The s3 path will change automatically, and I think in principal, if the dag were to change _so_ much as to break the .../org/task-org.json.gz it would break a lot of other things too, so I'd rather leave this for now, unless @jdu has any objections?

reach/airflow/dags/policy.py
nsorros: yep but this should be done independently outside of here to create the gold data and manually verified.

reach/airflow/tasks/evaluator.py
ivyleavedtoadflax: Do you mean extend the list with every jsonl that is yielded, rather than collecting in a list first? In practice these files are quite small, so i don't think memory efficiency is too much of a concern here. In fact I will abstract this away because it is already covered in the `_read_json_gz_from_s3` function

reach/airflow/dags/policy.py
ivyleavedtoadflax: Ahh yes, you're right. I did manually verify this step. hmmmmmmmm I didn't think of that. So at one level I would like to keep this step in airflow so that the EPMC matching of the gold data is done with exactly the same logic as the EPMC matching of the reach data. Perhaps it would make sense to:
* Create a separate DAG for doing this.
* Run that DAG independently, outputting a matched file which can then be verified and uploaded to S3
* Download that file from s3 in the policy DAG here, and run the final end-to-end evaluation of reach

reach/airflow/tasks/evaluator.py
jdu: If you look in policy.py in the DAGs folder we have `to_s3_output` and `to_s3_output_dir` which format a path based on some parameters (config, dag_id, etc..). At least changing this to use the to_s3_output method will mean if we make changes / refactor we can find the instances where it's used fairly easily.

ivyleavedtoadflax: this already uses a `to_s3_output_dir_short()` function which is based on `to_s3_output_dir()`

ivyleavedtoadflax:
```
    combineReachFuzzyMatchRefs = evaluator.CombineReachFuzzyMatchesOperator(
        task_id='EvaluateCombineReachFuzzyMatches',
        organisations=ORGANISATIONS,
        src_s3_dir_key=to_s3_output_dir_short(dag, 'fuzzy-matched-refs'),
        dst_s3_key=to_s3_output(
            dag, 'evaluation', 'fuzzy-matched-reach-refs', '.json.gz'),
        es_index='-'.join([dag.dag_id, 'epmc', 'metadata']),
        dag=dag,
    )
```

reach/airflow/dags/policy.py
nsorros: Ok but this independent DAG should run on an ad hoc basis since creation of the gold data happens irregularly compared to evaluating reach which should happen often

ivyleavedtoadflax: agreed

ivyleavedtoadflax: In https://github.com/wellcometrust/reach/pull/341/commits/75084fea759a969e9896d19dcce8ed8b1a2e29b5 I remove fuzzy match gold step to give the following dag for policy-test:

image

reach/airflow/tasks/evaluator.py
jdu: Re-reading this I don't really have a problem with the way it's set up here. I don't think we have any plans to change the semantics of how we store things in S3 and I think if it were to change it would require a refactor of areas across the board and not just here. So I wouldn't worry about how this is passed in right now.

reach/airflow/dags/policy.py
ivyleavedtoadflax: New matching dag implemented in https://github.com/wellcometrust/reach/pull/341/commits/195c58784ba25eb221e884e96b9a2f09d4cfc131

ivyleavedtoadflax: PR description updated

reach/airflow/dags/policy.py
```diff
@@ -190,14 +227,15 @@ def create_org_pipeline_fuzzy_match(dag, organisation, item_limits, spider_years
...
-    return parsePdf, org_fuzzy_match(
+    esIndexFuzzyMatched, fuzzyMatchRefs = org_fuzzy_match(
         dag, organisation, item_limits, parsePdf, esIndexPublications)
 
+    return parsePdf, esIndexFuzzyMatched, fuzzyMatchRefs
 
 def create_org_pipeline_exact_match(dag, organisation, item_limits,
                                     spider_years, parsePdf=None,
                                     esIndexFullTexts=None):
-    """ Creates all tasks needed for producing exact matched citatinos
```
lizgzil: I like the idea of calling citations 'citatinos' and 'citatinas'

ivyleavedtoadflax: :rofl: sounded rather spanish

reach/airflow/tasks/evaluator.py
ivyleavedtoadflax: great! :+1:

nsorros:
> EvaluateExtractRefsFromGoldData

why sudo?

nsorros:
> LGTM. I ran `sudo ./docker_exec.sh airflow test` for each of the EvaluateCombineReachFuzzyMatches, EvaluateExtractRefsFromGoldData, EvaluateFuzzyMatchGoldRefs and EvaluateResults tasks successfully, `make docker-test` also worked (4 passed, 1 xfailed, 4 warnings in 6.48s).

why sudo?

ivyleavedtoadflax:
I also always need to run sudo when running docker and docker-compose

nsorros:
> I also always need to run sudo when running docker and docker-compose

maybe for linux this is the solution https://docs.docker.com/install/linux/linux-postinstall/

ivyleavedtoadflax:
I don't think it's a real issue. It depends whether you want to grant docker those permissions.

nsorros:
> I don't think it's a real issue. It depends whether you want to grant docker those permissions.

true

PR #368, opened 2020-01-23
Fix issues in gold reference extraction (Fix #367)

### LINKED ISSUE #367: Fuzzy Matching with airflow does not produce the same results as running locally with reach_evaluator
Running the new `match-annotated-titles` dag produces fewer (or no matches) when compared to the fuzzy matcher run [locally](https://github.com/wellcometrust/datalabs/blob/master/analysis/reach_evaluator/scripts/match_manually_tagged_references.py) using input files:
s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid_TITLE.jsonl
s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid.jsonl.

Note this uses the branch in PR #368.
See [Makefile](https://gist.github.com/ivyleavedtoadflax/bfe0090a7f7922a75d041feeab172596) to speed up checking of airflow outputs.

ivyleavedtoadflax:
Fixes issues with the end to end evaluation introduce in #341, primarily documented in #367. The main changes this PR introduces are:

* Splits `EvaluateExtractRegsFromGoldData` into to separate tasks for easier debugging.
* References found in the manually tagged data that cannot be pinned to a document id will be discarded for the purposes of the end to end evaluation.
* Moved specification of elastic search parameters to the policy dag specification so that these can be passed to the `EvaluateOperator` and recorded for comparison with results.
* Solves most pressing issue with matches differing from local runs (#367) by creating a dummy `reference_id` based on a hash of the title. Future versions may allow the `reference_id` to be recreated in the annotation environment once other aspects of the reference have also been manually annotated..

## Type of change

- [x] :bug: Bug fix 
- [x] :sparkles: New feature

# How Has This Been Tested?

`match_annotated_titles` dag tested on local environment
`make docker-test`

# Checklist:

### DIFF
--- a/reach/airflow/dags/match_annotated_titles.py
+++ b/reach/airflow/dags/match_annotated_titles.py
@@ -17,8 +17,8 @@ DEFAULT_ARGS = {
     'retry_delay': datetime.timedelta(minutes=5),
 }
 
-REFERENCE_ANNOTATIONS="s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid_TITLE.jsonl.gz"
-TITLE_ANNOTATIONS = "s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid.jsonl.gz"
+TITLE_ANNOTATIONS="s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid_TITLE.jsonl.gz"
+REFERENCE_ANNOTATIONS = "s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid.jsonl.gz"
 
 # dag.id of dag which produced the EPMC ES Index
 ES_INDEX="policy"
--- a/reach/airflow/dags/match_annotated_titles.py
+++ b/reach/airflow/dags/match_annotated_titles.py
@@ -94,7 +94,7 @@ def create_match_dag(dag_id, default_args):
         refs_s3_key=REFERENCE_ANNOTATIONS,
         titles_s3_key=TITLE_ANNOTATIONS,
         dst_s3_key=to_s3_output(
-            dag, 'evaluation', 'extracted-gold-refs', '.json.gz'),
+            dag, 'extracted-gold-refs', '.json.gz'),
         dag=dag,
     )
 
@@ -104,7 +104,7 @@ def create_match_dag(dag_id, default_args):
         src_s3_key=extractedGoldRefs.dst_s3_key,
         organisation='gold',
         dst_s3_key=to_s3_output(
-            dag, 'evaluation', 'fuzzy-matched-gold-refs', '.json.gz'),
+            dag, 'fuzzy-matched-gold-refs', '.json.gz'),
         es_index='-'.join([ES_INDEX, 'epmc', 'metadata']),
         dag=dag,
     )
--- a/reach/airflow/dags/match_annotated_titles.py
+++ b/reach/airflow/dags/match_annotated_titles.py
@@ -49,18 +49,8 @@ def to_s3_output(dag, *args):
     slug = '-'.join(components)
     return (
         '{{ conf.get("core", "reach_s3_prefix") }}'
-        '/output/%s/%s/%s%s'
-    ) % (dag.dag_id, path, slug, suffix)
-
-
-def to_s3_output_dir(dag, *args):
-    """ Returns the S3 URL for any output directory for the DAG. """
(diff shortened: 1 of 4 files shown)


reach/airflow/dags/match_annotated_titles.py
```diff
@@ -86,12 +79,20 @@ def create_match_dag(dag_id, default_args):
         schedule_interval=None
     )
 
-    extractedGoldRefs = evaluator.ExtractRefsFromGoldDataOperator(
-        task_id='EvaluateExtractRefsFromGoldData',
+    matchedAnnotations = evaluator.AddDocidToTitleAnnotations(
+        task_id='EvaluateMatchAnnotations',
```
nsorros: should the task_id match with the name of the operator?

jdu: It should just for clarities sake, so that when we're looking at the DAG in airflow we can easily figure out which operator it's associated with.

reach/airflow/tasks/fuzzy_match_refs.py
```diff
@@ -144,14 +144,11 @@ class FuzzyMatchRefsOperator(BaseOperator):
...
-    SHOULD_MATCH_THRESHOLD = 80
-    SCORE_THRESHOLD = 50
-
     @apply_defaults
     def __init__(self, es_hosts, src_s3_key, dst_s3_key, es_index,
-                 score_threshold=SCORE_THRESHOLD,
+                 score_threshold=50,
                  organisation=None,
-                 should_match_threshold=SHOULD_MATCH_THRESHOLD,
+                 should_match_threshold=80,
```
nsorros: why hard coding?

reach/airflow/dags/match_annotated_titles.py
```diff
@@ -17,8 +17,11 @@
     'retry_delay': datetime.timedelta(minutes=5),
 }
 
-REFERENCE_ANNOTATIONS="s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid_TITLE.jsonl.gz"
-TITLE_ANNOTATIONS = "s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid.jsonl.gz"
+TITLE_ANNOTATIONS="s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid_TITLE.jsonl.gz"
+REFERENCE_ANNOTATIONS = "s3://datalabs-data/reach_evaluation/data/sync/2019.10.8_valid.jsonl.gz"
+
+# dag.id of dag which produced the EPMC ES Index
+ES_INDEX="policy"
```
nsorros: should this be somewhere else or can you retrieve it from somewhere?

ivyleavedtoadflax: Maybe, yes. It's here because it's better if this refers to the index that is built in the policy dag rather than rebuilding it here itself.

reach/airflow/dags/match_annotated_titles.py
ivyleavedtoadflax: Agree

reach/airflow/tasks/fuzzy_match_refs.py
ivyleavedtoadflax: This is now defined in the dag rather than the task definition, so to remove confusion about where this actually happens, I didn't want to leave the two variables here, because the same two are defined in the dag.

jdu:
The location of the gold data under the `datalabs-data` bucket is inaccessible to airflow from staging and prod. Can you please move copies of the data to datalabs-<env>/airflow-reach/data/<filename> for each environment (-dev, -staging, -prod) and update the paths above accordingly?

This would have worked locally for you as you have access to datalabs-data through your AWS creds, but the airflow workers creds are more locked down than our individual creds and `datalabs-data` is not a public bucket.

ivyleavedtoadflax:
Have added https://github.com/wellcometrust/datalabs-infra/pull/345 to propagate the `reach_s3_data_prefix` into `staging`. Will make changes in reach to address this in `dev`.

ivyleavedtoadflax:
> Have added [wellcometrust/datalabs-infra#345](https://github.com/wellcometrust/datalabs-infra/pull/345) to propagate the `reach_s3_data_prefix` into `staging`. Will make changes in reach to address this in `dev`.

Actually wasn't required.

PR #372, opened 2020-01-27
Update airflow to 1.10.7 (and other deps)

SamDepardieu:
# Description

Bumps dependency versions in our requirements file.
Also fix a small bug where Scrapy's logger didn't properly send colors to its logs, causing Airflow's logger to send warnings.

## Type of change

- [ ] :sparkles: New feature
- [ ] :fire: Breaking change

# How Has This Been Tested?
Local runs of `policy-test` dag
`make docker-test`


requirements.txt
```diff
@@ -1,130 +1,133 @@
 # Created by 'make update-requirements-txt'. DO NOT EDIT!
```
jdu: We need to be a bit more careful about bulk updating dependency versions like this I think. Some of these updates are major version updates which could cause breaking changes or unintended changes to how our code works.

SamDepardieu: Yeah, I'm not very happy with how this make update_requiorements_txt works... But can't think of something better that wouldn't require us to check line by line that everything is up to date

jdu: I think maybe we just pin the versions and get rid of the automated update-requirements stuff. Once we're in public release mode we shouldn't need to update these very often. 

I'm just worried that at some point we're going to be prepping a release and we do one of these updates throwing us into a situation where we're trying to scramble to figure out which library upgrade broke it and downgrading things... we don't need to be on the constant bleeding edge of the version for all these dependencies really.

Issue #343, opened 2020-01-17
LXML Errors in pdf parser


jdu:
Some XML parsing errors are being suppressed in the PDF parser which is resulting in PDF policy documents not being included in the overall doc sets for various sources.

Issue has been traced to LXML being very strict and not allowing specific characters in attribute values unless they are explicitly escaped.

kristinenielsen:
@TeriForey this might be one for you to pick up if you have some free time

jdu:
This is fixed by #376

PR #376, opened 2020-01-30
Various fixes for the following:

jdu:
# Description

Fixes for #343 

- Remove JOBDIR so that scrapers perform a full / fresh scrape on each run
- Remove the dont_filter key config so that scrapers properly avoid crawling known pages
- Added a function to get the DocumentID from a given PDF to use as an identifier (disabled)   -
- Fixes an issue with LXML parsing where the output from pdftohtml was not proper XML
- Changed the manifest update code to update a hash entry regardless of whether it's existing in order to support changes to data schemas

## Type of change

- [ ] :sparkles: New feature
- [ ] :fire: Breaking change

# How Has This Been Tested?

Ran policy-test and validated results

Ran `make docker-test`

# Checklist:

- [x ] New and existing unit tests pass locally with my changes


reach/scraper/wsf_scraping/file_system.py
```diff
@@ -4,12 +4,24 @@
...
+import re
+
+from lxml import etree
 from abc import ABC, abstractmethod
 from botocore.exceptions import ClientError
 
 
 def get_file_hash(file_path):
-    """Return the md5 hash of a file."""
+    """ Return both possible IDs for a given document
```
ivyleavedtoadflax: Is this docstring right? It looks like it is only returning `<file hash>`?

ivyleavedtoadflax: AH ok this is commented out. Actually since we are going to do a new complete run, it would be very interesting to have both ids, to be able to compare them.

Issue #500, opened 2020-04-22
Incorporate new multitask deep reference parser model


lizgzil:
Working on in https://github.com/wellcometrust/reach/pull/505

Will involve changes to `yield_structured_references` in `refparse.py` most likely.

- Class name changes
- Output format changes
- Will need to point to github location instead of S3 for the .whl download in Makefile

PR #505, opened 2020-04-27
Integrate multitask splitter+parser model

lizgzil:
# Description
Note: No need to review anything from `reach/` as this will all be deleted after the merge.

Addresses issue https://github.com/wellcometrust/reach/issues/500.

In this PR I replace `SplitSection` with `SplitParser` in `refparse.py`. This means instead of using the deep reference splitter model plus the naive bayes parser model, we now combine splitting and parsing by using the deep reference multitask model.

This has knock-on effects for various other scripts - changing the tests, the init and settings files, the functions in `parse.py` (now reduced to just one), adding a file (test_config_multitask.ini) which will make the tests on this model a bit quicker than they would be with the default config file.

**Note**: I make changes to both the scripts in `reach/refparse` and `split_reach/extracter/refparse`, which are essentially the same changes for both file locations. In each commit I add each pair of scripts (e.g. `reach/refparse/utils/__init__.py` and `split_reach/extracter/refparse/utils/__init__.py`) so in terms of reviewing more easily you might want to go through commit by commit.

## Type of change

- [x] :sparkles: New feature
- [ ] :fire: Breaking change

# How Has This Been Tested?

`make test`

# Checklist:


reach/refparse/refparse.py
```diff
@@ -144,33 +137,27 @@ def get_file(
...
 
-    # Loading the pipeline
-    model = get_file(model_file, 'pickle')
-
     sectioned_documents = transform_scraper_file(scraper_file)
 
     # Instantiate deep_reference_parser model here (not in loop!)
-
-    section_splitter = SplitSection()
+    splitter_parser = SplitParser(config_file=MULTITASK_CFG)
```
nsorros, line 160: possibly the config_file should be the default on the version you shipped to not expose that here but this is a minor thing

reach/refparse/utils/__init__.py
```diff
@@ -1,13 +1,10 @@
...
+from .parse import structure_reference
 from .fuzzy_match import FuzzyMatcher
 from .file_manager import FileManager
 from .serialiser import serialise_matched_reference, serialise_reference
 from .exact_match import ExactMatcher
 
 __all__ = [
-    split_reference,
-    predict_components,
-    merge_components,
```
nsorros, line 10: don't you still need to do merge?

split_reach/extracter/refparse/refparse.py
```diff
@@ -21,20 +21,15 @@
                    ExactMatcher)
 from .settings import settings
 
-from deep_reference_parser.split_section import SplitSection
+from deep_reference_parser.split_parse import SplitParser
+from deep_reference_parser.common import MULTITASK_CFG
```
nsorros, line 25: is this duplciation? why there are two files doing the same thing?

reach/refparse/refparse.py
```diff
@@ -179,23 +166,18 @@ def yield_structured_references(scraper_file,
...
 
         logger.info('[+] Extracted {} references from document {}'.format(
-            len(splitted_references),
+            len(reference_predictions),
             i
         ))
 
-        # For some weird reason not using pool map
-        #   in my laptop is more performant
-        structured_references = pool_map(
```
nsorros, line 193: this is an abstraction so that you can parallelise the work across cores. how many workers did you use? check the num_workers in your laptop, setup to None to utilise all cores and check again. even then i don't think you should remove this as we need to investigate why using more cores performs worse, this does not make sense.

nsorros:
This PR moves into a combined split / parse logic but note that this is not necessary. The multi task model does not make joint predictions, just trains on two tasks i.e. multitask. You can still have a separate split and parse method which would make integrating a bit easier.

reach/refparse/tests/test_config_multitask.ini
```diff
@@ -0,0 +1,39 @@
+[DEFAULT]
```
nsorros, line 1: this should not be here.

reach/refparse/utils/parse.py
```diff
@@ -4,121 +4,23 @@
...
-
-
-def is_year(component):
-    if len(component) == 6:
-        component = component[1:5]
-    return component.isdecimal() and VALID_YEARS[0] <= int(component) < VALID_YEARS[1]
-
-
-def predict_components(model, reference_components):
+def structure_reference(reference_components):
```
nsorros, line 7: is that doing the job of merge_components?

requirements.txt
```diff
@@ -164,5 +164,6 @@ zope.deprecation==4.4.0
 zope.interface==4.7.1
 https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-2.1.0/en_core_web_sm-2.1.0.tar.gz
 https://github.com/keras-team/keras-contrib/tarball/master
-https://datalabs-public.s3.eu-west-2.amazonaws.com/deep_reference_parser/deep_reference_parser-2019.12.1-py3-none-any.whl
 https://datalabs-public.s3.eu-west-2.amazonaws.com/reach_evaluator/reach_evaluator-2020.1.1-py3-none-any.whl
+
+https://github.com/wellcometrust/deep_reference_parser/releases/download/2020.4.29/deep_reference_parser-2020.4.5-py3-none-any.whl
```
SamDepardieu, line 169: This line needs to be updated in `split_reach/extracter/requirements.txt` too :)

lizgzil: oh i didnt realise each component had its own makefile+requirements!

reach/refparse/utils/__init__.py
lizgzil: no, everything is done all together in `parse.structure_reference` now.

reach/refparse/tests/test_config_multitask.ini
lizgzil: we need the test config file for `test_split_parse.py` since the default `config_multitask.ini` file points to an unnecessarily large embeddings file which would slow down the testing significantly. Or do you mean it should be located in a different folder?

reach/refparse/utils/parse.py
lizgzil: yep

reach/refparse/refparse.py
lizgzil: For reach we are just setting `pool_map = map` so no parallelisation. I will add an issue to investigate adding a proper pool_map and having a suitable number of workers, but I think it's out of scope for this PR

requirements.txt
lizgzil: https://github.com/wellcometrust/reach/pull/505/commits/0ffba4f368dd736bb28e30bec6adf70932ee5394

lizgzil:
@nsorros 👍 for the review, on your points:
- "having a test config file here" - see my response above, I was a bit unsure what you meant
- "combining split and parse which is not strictly necessary" - As we discussed the new multitask model does indeed make predictions of both tasks, so it is necessary to combine.
- "remove pool_map which should be avoided or better understood" : I've added an issue https://github.com/wellcometrust/reach/issues/507, but for now the pool_map isn't used for multiprocessing since `pool_map = map`, so I hope it's ok to address this at another point.

reach/refparse/refparse.py
nsorros: why do you say that in reach we set pool_map = map?

nsorros: ok i see, i was under the impression that locally it was as you said but in prod it was using multiprocessing but i think you are right. @SamDepardieu or @jdu can confirm. regardless removing pool_map will make finding a way to parallelise in the future a difficult exercise for sam or jeff, out of memory sam spent one week profiling to be able to place the parallelisation there. so in any case it is worth consulting this with the devs.

reach/refparse/tests/test_config_multitask.ini
nsorros: i meant that it should live in deep reference parser not in reach. also if it ends up staying here maybe you can remove everything but the build part as the others are not needed.

nsorros:
> @nsorros 👍 for the review, on your points:
> 
> * "having a test config file here" - see my response above, I was a bit unsure what you meant
> * "combining split and parse which is not strictly necessary" - As we discussed the new multitask model does indeed make predictions of both tasks, so it is necessary to combine.
> * "remove pool_map which should be avoided or better understood" : I've added an issue #507, but for now the pool_map isn't used for multiprocessing since `pool_map = map`, so I hope it's ok to address this at another point.

ideally the test file should be in the deep reference end not here. reach should use deep reference parser as a package without needing to define such a config to test.

split and parse could still be used separately, this would mean predicting twice, it would make integrating easier but i understand this is inefficient, just making the point that an easier transition could be made before combining the two.

removing pool_map should be left to devs, it is up to them but it is not as simple as saying that map is used at the moment, as you are breaking an easy way to parallelise the code in case running reach takes longer than needed.

lizgzil:
@nsorros good point about the testing! Of course it should just be tested in the deep reference parser repo. I've removed these tests and the config (from `split_reach/` only since `reach/` will be deleted anyway) in this commit https://github.com/wellcometrust/reach/pull/505/commits/c623c5b2f378714bb7cf1e8cdaa5e470a8e32512

nsorros:
> I'm happy with the requirement change, but I'll leave the final approval to @nsorros

could you comment on the removal of pool_map as well?

reach/refparse/refparse.py
lizgzil: @SamDepardieu and @jdu could you comment on nick's comment about `pool_map`? As far as i see we have it, but don't use its parallelisation abilities in Reach ever, is that correct? Do you think we need to change anything with that?

This is the only thing standing in the way of this PR getting merged.

SamDepardieu:
@nsorros Sorry, just saw your comment. It's okay to remove it for the time being, but it'd probably be good to put it back at some point, to speed things up. In the first time, we can monitor this to see if it actually needs this

reach/refparse/refparse.py
```diff
@@ -144,33 +137,27 @@ def get_file(
 
 
 def yield_structured_references(scraper_file,
-                pool_map, logger, model_file=DEFAULT_MODEL_FILE):
+                pool_map, logger):
```
nsorros, line 140: no need to pass pool_map here

reach/refparse/refparse.py
```diff
@@ -144,33 +137,27 @@ def get_file(
...
-            Defaults to model within our source tree.
     """
 
     logger.info("[+] Reading input files")
 
     # Loading the scraper results
     scraper_file = get_file(scraper_file, "", get_scraped=True)
 
-    # Loading the pipeline
-    model = get_file(model_file, 'pickle')
```
nsorros, line 167: i know sam hated this function get_file but i still remember it with nostalgia

reach/refparse/refparse.py
```diff
@@ -224,14 +205,13 @@ def yield_structured_references(scraper_file,
     )
 
 
-def parse_references(scraper_file, model_file, num_workers, logger):
+def parse_references(scraper_file, num_workers, logger):
```
nsorros, line 208: you also do not need num_workers anymore

reach/refparse/utils/parse.py
nsorros: add input output or args and returns statements

reach/refparse/utils/parse.py
```diff
@@ -4,121 +4,23 @@
...
-    predicted_components = predict_components(
-        model,
-        splitted_components
-    )
+    # Keep empty strings for classes not predicted from the deep reference parser
+    # Useful for possible downstream errors
+    structured_reference = {ref_class: '' for ref_class in settings.REF_CLASSES}
+    for component in settings.DRP_REF_COMPONENTS:
+        structured_reference[
+            settings.COMPONENT_NAME_MAP.get(component, component)
```
nsorros, line 23: i would add a comment in the end of the line to say "leave component unchanged if no map" as the logic `.get(component, component)` makes you stop and think what is going on

reach/refparse/utils/parse.py
```diff
@@ -4,121 +4,23 @@
...
+            ] = ' '.join([r[0] for r in reference_components if r[1]==component])
```
nsorros, line 24: this logic of merge is different from the previous one. here you are merging all components that have been tagged the same. i understand that we do not have a way to evaluate the impact of this but could we write an email to @jdu to push about the resolution of the unique id or get a timeline for it, it is quite important to be able to argue those changes.

split_reach/extracter/refparse/utils/parse.py
```diff
@@ -4,121 +4,23 @@
...
+    ref_components = [r[1] for r in reference_components]
 
-    structured_reference = merge_components(predicted_components)
+    # Keep empty strings for classes not predicted from the deep reference parser
+    # Useful for possible downstream errors
+    structured_reference = {ref_class: '' for ref_class in settings.REF_CLASSES}
+    for component in settings.DRP_REF_COMPONENTS:
+        structured_reference[
+            settings.COMPONENT_NAME_MAP.get(component, component)
+            ] = ' '.join([r[0] for r in reference_components if r[1]==component])
```
nsorros, line 24: i find it really difficult to repeat comments and i am afraid that if you keep both changes that code in the two different places will differ. i will not hold the changes due to that but you have been warned :P

reach/refparse/refparse.py
lizgzil: this file will be deleted (it's in `reach/`)

reach/refparse/refparse.py
lizgzil: again, will be deleted

reach/refparse/utils/parse.py
lizgzil: ok will do, but I'm just going to make the change to `split_reach/extracter/refparse/utils/parse.py`

reach/refparse/utils/parse.py
lizgzil: The reason for this is that the deep reference parser doesn't output probabilities of the predictions, which was the information needed for the old way of finding the most likely 'group' of authors, titles, etc. Perhaps there is a better way to deal with this though than just grouping them all together

split_reach/extracter/refparse/utils/parse.py
lizgzil: Sam is just going to delete the entire `reach/` folder, so I'm only making the edits to the `split_reach` folder now, but I am being careful. It was a bit of a nightmare to undo the commits since they were all paired.

PR #520, opened 2020-05-18
Fix fuzzymatcher task

SamDepardieu:
# Description

Reach fuzzy-matcher task had a few issues:
 - No defaults for `should_match_threshold` and `score_threshold`, leading to some issue with the ES query
 - Fix a typo in the dict index we use to store `associated_policy_docs`

## Type of change

# How Has This Been Tested?
Local runs
Runs on staging argo
`make docker-test`

### DIFF
--- a/pipeline/reach-fuzzy-matcher/fuzzymatcher_task.py
+++ b/pipeline/reach-fuzzy-matcher/fuzzymatcher_task.py
@@ -188,7 +188,7 @@ class FuzzyMatchRefsOperator(object):
                 ref_id = fuzzy_matched_reference['reference_id']
                 if ref_id in references.keys():
 
-                    references[ref_id]['policies'][
+                    references[ref_id][
                         'associated_policies_count'] += 1
 
                     references[ref_id]['policies'].append(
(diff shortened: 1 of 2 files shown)


pipeline/reach-fuzzy-matcher/fuzzymatcher_task.py
```diff
@@ -255,12 +255,12 @@ def execute(self):
     )
     arg_parser.add_argument(
         '--score_threshold',
-        default=None,
+        default=50,
```
lizgzil, line 258: I can't remember whether 50 was the threshold for the old method of fuzzy matching metric, and whether this number would be suitable for an elastic search match metric? Any memory of having chosen the 50 threshold?

SamDepardieu: I can only gurarantee it was also 50 in the old version (You can check on the release branch, still running Airflow etc -> https://github.com/wellcometrust/reach/blob/release/reach/airflow/dags/policy.py#L41 ).

As to why, I think this actually is a question for @nsorros

PR #598, opened 2020-08-25
Fix citations not grouping together

### LINKED ISSUE #596: Publications cited in more than one policy document are not grouped together
Publications that appear in more than one policy document are not grouped together. 

Feedback has shown that users think the policy citations are duplicates and lose trust in the product.

SamDepardieu:
# Description

Citations were grouped by extracted reference id. Some of the extracted references have slightly different titles
because of spacing, casing or fuzzymatching. We are now using the matched reference id instead to group the citations.

Fix #596 

## Type of change

# How Has This Been Tested?

`make docker-test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -228,7 +228,7 @@ run-fuzzymatcher: fuzzymatcher-image
 		-e ES_HOST=${DOCKER_LOCALHOST} \
 		--network=host \
 		${ECR_ARN}/reach-fuzzy-matcher \
-		${EXTRACTER_PARSED_DST} \
+		${EXTRACTER_SPLIT_DST} \
 		${FUZZYMATCHER_DST} \
 		${ORG} \
 		epmc_metadata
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -228,7 +228,7 @@ run-fuzzymatcher: fuzzymatcher-image
 		-e ES_HOST=${DOCKER_LOCALHOST} \
 		--network=host \
 		${ECR_ARN}/reach-fuzzy-matcher \
-		${EXTRACTER_PARSED_DST} \
+		${EXTRACTER_SPLIT_DST} \
```
lizgzil, line 231: I'm not sure if this should change? should be same input right?

SamDepardieu: Nope, the input here was wrong. The `EXTRACTER_PARSED_DST` is only the extracted unorganised references, whch can't be processed by the current version of the fuzzymatcher. The fuzzymatcher is expected the formatted version in `EXTRACTER_SPLIT_DST`. 
An easy way to tell is that the task doesn't return anything with the `PARSED_DST` one 👍

lizgzil: ah cool!

lizgzil (approved):
looks good 👍

